PR #3, opened 2013-11-21
Fixed priority inversion by removing the boolean flag in RosPubChannelElement and added unbuffered publishing

meyerj:
There was a priority inversion in the `RosPubChannelElement` as a real-time thread that called `signal()` had to wait until the non-real-time publish activity has finished it's loop due to the mutex in `RosPublishActivity` (9cb131209f9de887999fd85439ed487e9fd00ad6). As discussed with Ruben today, the boolean flag is not required at all as just calling `publish()` on each publisher is not expensive if there are no new elements in the queue (a4e8b961895e6de88a8f281de76c64cc7aa76778).

The third commit moves the potentially large data object containing vectors etc. from the stack of the `publish()` method to a member variable. This should reduce the number of reallocations if subsequent messages contain vectors of the same size (c210eb33dc38527b2e3d45264f47da28da291363).

Last but not least, I added support for unbuffered publishers, where the thread writing in the port directly calls the ROS publisher's `publish()` method. This is not real-time safe under certain conditions, but there are certainly use cases (e.g. when publishing from a non-real-time thread or for simulation). Note that you still can specify a buffer size for roscpp's internal buffer mechanism by increasing the `size` property of the `ConnPolicy`. I also added a `ros.topicUnbuffered(topic)` operation to the rtt_ros service (888e9ffa76d2e92fe01a56b9e11bc3f13b97d19b).

By the way, shouldn't there be `size` argument even for data and unbuffered variants of the `ros.topic*(topic[,size])` operation, as size is always passed as `queue_size` paramter to the ROS publisher? Do operations support optional arguments to not break existing scripts?


psoetens:
I just reviewed it, couldn't spot a problem/potential issue with the new code. Worth merging & further testing.

doudou:
Hey why don't you use the RTT Data / Buffer constants for the connection types ?

meyerj:
> Hey why don't you use the RTT Data / Buffer constants for the connection types ?

Because there is no constant for unbuffered connections in the `RTT::ConnPolicy` class yet. Of course we could add it, but for port connections that does not make sense as you always need a data or buffer element in between and rtt_ros_integration would depend on a newer version of RTT. The unbuffered publisher forwards written data elements directly to the ROS publisher from the thread of the writer without involving the publish activity. It's different from publishing with `DATA`.

doudou:
aaaah. Missed the difference

I would definitely +1 the addition of such a policy type in RTT. This would
be useful for corba and MQ as well (all transports now have at least one
thread between the component(s) and the transport)

2013/11/22 Johannes Meyer notifications@github.com

psoetens:
On Fri, Nov 22, 2013 at 9:30 AM, Sylvain Joyeux notifications@github.comwrote:

> aaaah. Missed the difference
> 
> I would definitely +1 the addition of such a policy type in RTT. This
> would
> be useful for corba and MQ as well (all transports now have at least one
> thread between the component(s) and the transport)

+1. Please let's then standardise the buffer policy number... Does -1 make
sense ?

Peter

meyerj:
We could add such a constant to RTT 2.7 and use it in the [hydro-devel](https://github.com/orocos-toolchain/rtt_ros_integration/tree/hydro-devel) branch of rtt_ros_integration, where the latest updates in master are going to be merged.

meyerj:
> +1. Please let's then standardise the buffer policy number... Does -1 make
> sense ?

I took -1 as it is kind of special and not handled by `internal::ConnFactory::buildDataStorage<T>(policy)` (returns an uninitialized shared_ptr). Do you have any preference?

doudou:
So far, they were positive numbers, so I guess we should take the next one
in line, or ?

2013/11/22 Peter Soetens notifications@github.com

psoetens:
On Fri, Nov 22, 2013 at 1:00 PM, Sylvain Joyeux
notifications@github.com wrote:

> So far, they were positive numbers, so I guess we should take the next one
> in line, or ?

This was also my initial gut feeling, but -1 is kind-of a special case
where you say 'nothing/none'.

Peter

PR #5, opened 2013-12-04
rtt_roscomm: separated typekit pkg template from source file templates required to build the typekit

meyerj:
The typekits and service proxies will be built in a separate subdirectory now and do not influence the scope of the caller. I think this is much cleaner now, as with the previous version, just adding `find_package(rtt_roscomm)` to your `CMakeLists.txt` set the `CMAKE_BUILD_TYPE` to `MinSizeRel` and using the macros also affected other targets built in the caller's list file.

The contents of the former `GenerateRTTROSCommPackage.cmake.em` and the new `CMakeLists.txt` files in `src/templates` are basically the same. The macro argument `package` will be stored in the temporary `_package` variable before the macro builds the typekit/service proxy from the source directory in `src/templates`. I will add two other minor changes as commit annotations (see below).

This PR includes https://github.com/jhu-lcsr-forks/rtt_ros_integration/pull/55 (rtt_roscomm: Unified naming scheme for rtt_rosservice and use a singleton for the ROSServiceRegistryService), as otherwise they would be difficult to merge.


smits:
Loods good for me.

PR #6, opened 2013-12-09
Merge rtt_ros_comm and rtt_common_msgs into rtt_ros_integration?

meyerj:
As rtt_actionlib and the test packages in [rtt_ros_integration](https://github.com/orocos-toolchain/rtt_ros_integration/tree/hydro-devel) depend on the typekit packages in the [rtt_ros_comm](https://github.com/orocos-toolchain/rtt_ros_comm/tree/hydro-devel) and [rtt_common_msgs](https://github.com/orocos-toolchain/rtt_common_msgs/tree/hydro-devel), they should be integrated in the rtt_ros_integration repository and always released as a unit with the same version number. With catkin and bloom there is no reason to keep them separated anymore and the existing meta-packages for rtt_common_msgs and rtt_ros_comm will still be available.

Merging was already proposed by @jbohren in https://github.com/jhu-lcsr-forks/rtt_ros_integration/issues/44.

Which directory structure do you prefer for the merged repository?
- Merge them as-is in the root of the repository (all packages at the same level)
- Merge them in the `rtt_common_msgs` and `rtt_ros_comm` subdirectories of the [hydro-devel](https://github.com/orocos-toolchain/rtt_ros_integration/tree/hydro-devel) tree in rtt_ros_integration (Note that there will be two similarly named folders `rtt_roscomm` and `rtt_ros_comm` then.)
- Merge them in the `common_msgs` and `ros_comm` subdirectories
- Additionally move the original rtt_ros_integration packages in a `rtt_ros_integration` folder?
- ...

I am not sure if rtt_geometry should also be merged in. These packages have some additional dependencies that might not be available for source builds. The resulting binary packages can still be installed independently.

Currently the [devel-hydro-rtt_ros_integration](http://jenkins.ros.org/job/devel-hydro-rtt_ros_integration/ARCH_PARAM=amd64,UBUNTU_PARAM=precise,label=devel/) job on Jenkins fails because of the missing dependency rtt_actionlib_msgs.


jbohren:
Maybe since they're really just typekits, instead of `rtt_ros_comm` we should rename this as `ros_typekits`?

smits:
I like that suggestion, but in this case I would add the rtt_common_msgs in
that subdir too.

Ruben

On Mon, Dec 9, 2013 at 6:38 PM, Jonathan Bohren notifications@github.comwrote:

> Maybe since they're really just typekits, instead of rtt_ros_comm we
> should rename this as ros_typekits?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/orocos-toolchain/rtt_ros_integration/issues/6#issuecomment-30153462
> .

## 

Ruben Smits, CTO
+32 479 511 786
Intermodalics - Kapeldreef 60, 3001 Heverlee - BELGIUM
www.intermodalics.eu

---

This email and any attached files are confidential and may be legally
privileged. Any copy, print or forward of this email, without the agreement
of sender or addressee, is strictly prohibited. Misuse is a violation of
the law on personal data protection (D. Lgs. 196/2003) and on secrecy of
correspondence (art. 616 cp). If you have received this transmission in
error please notify the sender immediately and then delete this email and
any attached files.

jbohren:
> I like that suggestion, but in this case I would add the rtt_common_msgs in that subdir too.

+1

meyerj:
> Maybe since they're really just typekits, instead of rtt_ros_comm we should rename this as ros_typekits?

I prefer just `typekits` instead of `ros_typekits` and not to introduce a new prefix?

This subdirectory will contain all typekit packages in rtt_ros_comm and rtt_common_msgs **and** the rtt_ros_comm and rtt_common_msgs meta-packages, without separating them in additional subdirectories, right?

Something like (+ using git-filter-branch to keep the history):

```
mkdir rtt_ros_integration/typekits/
mv rtt_ros_comm/* rtt_ros_integration/typekits/
mv rtt_common_msgs/* rtt_ros_integration/typekits/
```

Also the new service proxy packages (like rtt_std_srvs) will be in that folder although they do not provide typekits (yet), but I think this would be okay.

jbohren:
> This subdirectory will contain all typekit packages in rtt_ros_comm and rtt_common_msgs and the rtt_ros_comm and rtt_common_msgs meta-packages, without separating them in additional subdirectories, right?

Right, that does make more sense.

> Also the new service proxy packages (like rtt_std_srvs) will be in that folder although they do not provide typekits (yet), but I think this would be okay.

Yeah, the goal is to have typekits for the .srv types in the future, as well.

smits:
+1

PR #7, opened 2013-12-10
rtt_rosparam: Added support for subservices

meyerj:
Moved from https://github.com/jhu-lcsr-forks/rtt_ros_integration/pull/50 (with one additional patch):

This patch will add sub-service support to rtt_rosparam. A component could have nested services that also have properties. Currently it is not possible to load or store parameters of nested services.

There are several modifications, that I shortly summarize here:
1. I added policy-specific operations `getAllXXX()` and `setAllXXX()` in analogy to the per-parameter operations. This is not directly related to services, but I need it to fetch all parameters of a component with a nested service from another namespace than it's private component namespace (dd2ef57).
2. The `getAllXXX()` and `setAllXXX()` operations will recurse into services and also get/set their properties from/on the parameter server. This will only work for services, not for peers.
3. If a service is passed to `getXXX(name)` or `setXXX(name)` operations and no property of that name exists, the rosparam service will search a service with that name and recursively call `getAllXXX()`/`setAllXXX()` for that specific service only.

Note that `getAllAbsolute()` and `setAllAbsolute()` (where the latter probably does not make so much sense anyway) currently are broken due to https://github.com/ros/ros_comm/pull/313.

### DIFF
--- a/rtt_rosparam/README.md
+++ b/rtt_rosparam/README.md
@@ -29,16 +29,31 @@ policy (see below).
 
 #### Operations (getting/setting params)
 
-##### Operations on all properties
-* **getAll()** Attempt to get all properties from ros parameters in the
-  **COMPONENT** namespace.
-* **setAll()** Set ROS parameters in the **COMPONENT** namespace from all
-  properties of this component.
+##### Operations for getting all properties
+* **getAll()** or **getAllComponentPrivate()** Attempt to get all properties of this component (and its sub-services)
+  from the ROS parameter server in the **COMPONENT** namespace.
+* **getAllRelative()** Attempt to get all properties of this component (and its sub-services)
+  from the ROS parameter server in the relative namespace.
+* **getAllAbsolute()** Attempt to get all properties of this component (and its sub-services)
+  from the ROS parameter server in the absolute namespace.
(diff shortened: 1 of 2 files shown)


jbohren:
Should this be merged here or on gitorious?

meyerj:
This could be merged here. Only the toolchain packages (log4cpp, rtt, ocl, ...) are officially hosted at [Gitorious](https://www.gitorious.org/orocos-toolchain) and [GitHub](https://github.com/orocos-toolchain) is only a mirror.

We discussed to move the the rtt_ros_integration repositories to the [orocos](https://github.com/orocos) organization to separate them clearly, but I am not sure how many people are already using them from here. Originally they also have been mirrors of http://git.mech.kuleuven.be/ and most references in the web still point to this address.

jbohren:
> We discussed to move the the rtt_ros_integration repositories to the orocos organization to separate them clearly, but I am not sure how many people are already using them from here.

Yeah, I think that could make more sense, to emphasize that these tools depend on the toolchain, but aren't a part of the core.

So I think after merging https://github.com/orocos-toolchain/rtt_ros_integration/pull/6 it makes sense that we could move rtt_ros_integration over to the [orocos](https://github.com/orocos) organization.

PR #9, opened 2013-12-16
Added more types to rtt_rosparam

meyerj:
This pull request adds support for the basic types `unsigned int`, `char`, `unsigned char` and the corresponding vector types to rtt_rosparam.

The second patch adds property decomposition of unknown types to support properties of complex types like from kdl_typekit or eigen_typekit. This replaces the deprecated `PropertyBagIntrospection` that was present in groovy's [rosparam service](https://github.com/orocos/rtt_ros_integration/blob/groovy-devel/rtt_rosnode/src/rtt_rosparam_service.cpp).

Note that using property decomposition even types that have vector characteristics (like eigen_vector) are stored as structs and can only be loaded as structs on/from the ROS parameter server. I think it would be a nice add-on to somehow "detect" vector types (e.g. based on the presence of a `size` and `capacity` members in the decomposed PropertyBag) and store and load them as XmlRpc arrays. For other types like `KDL.Vector` it is not clear whether they should be stored as structs or arrays, but at least loading should support both variants.

### DIFF
--- a/rtt_rosparam/src/rtt_rosparam_service.cpp
+++ b/rtt_rosparam/src/rtt_rosparam_service.cpp
@@ -1,6 +1,7 @@
 #include <rtt/RTT.hpp>
 #include <rtt/Property.hpp>
 #include <rtt/plugin/ServicePlugin.hpp>
+#include <rtt/types/PropertyDecomposition.hpp>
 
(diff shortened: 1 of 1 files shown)


smits:
Are there any unit-tests pending?

PR #10, opened 2014-01-06
Cleanup before release (adding changelogs, moving unimplemented packages to separate branches)

jbohren:

PR #11, opened 2014-01-06
Move ROS-dependent packages from rtt_geometry to rtt_ros_integration

jbohren:
This PR includes changes from https://github.com/orocos/rtt_ros_integration/pull/10 and adds rtt_tf to rtt_ros_integration so that rtt_geometry can only depend on the orocos toolchain and KDL.

Note that it is removed from rtt_geometry in https://github.com/orocos/rtt_geometry/pull/8.


smits:
Why did you not move rtt_kdl_conversions?

jbohren:
So kdl_lua seems to depend on rtt_kdl_conversions. It doesn't look like it needs to, but I wanted to check first. I'm not too familiar with how the lua stuff is used, though. Are there tests that use it anywhere?

jbohren:
@smits So the following in kdlutils.lua uses some stuff from rtt_kdl_conversions, which means the only packages in rtt_geometry which currently don't depend on ROS libraries are kdl_typekit and eigen_typekit. 

``` lua
function frame_to_msg(f)
   msg = rtt.Variable("geometry_msgs/Pose")
   rtt.provides("KDL"):FrameToMsg(f,msg)
   return msg
end
function msg_to_frame(msg)
   f = rtt.Variable("KDL/Frame")
   rtt.provides("KDL"):MsgToFrame(msg,f)
   return f
end
```

smits:
Ok, I merged orocos/rtt_geometry#9 so we can safely move rtt_kdl_conversions to rtt_ros_integration

meyerj:
Merged in #13.

PR #19, opened 2014-01-15
Numerous bugfixes and enhancements in rtt_rosparam and rtt_rossservice

jbohren:
Also adding convenient way to push data from ROS parameter server to Eigen VectorXd types without the "1: 2: 3: etc..." struct format.


jbohren:
@smits Any chance these changes could be merged?

rtt_roscomm/README.md
```diff
@@ -88,6 +92,29 @@ stream("my_component.my_output", rostopic.connection("my_ros_output"))
...
+You can also set up these connections in C++ code:
+```cpp
+
+#include <rtt_rostopic/rostopic.h>
+#include <rtt/internal/GlobalService.hpp>
+
+// ...
+
+rtt_rostopic::ROSTopic rostopic(NULL);
+rostopic.connectTo(RTT::internal::GlobalService::Instance()->provides("rostopic"));
```
psoetens: Is this connectTo() step necessary ? Is there a use case where you want it to connect to something else than the rostopic service in the global service ?

psoetens:
I think we should merge it, but @smits and @jbohren might take a closer look at the rosparam changes.

rtt_roscomm/README.md
jbohren: It's not necessary, this section was just showing how to do it in C++ code if you wanted to. This can be useful for certain topics that are meant to only publish to ROS (visualization markers, for example).

psoetens: Well, I mean, why isn't it just:

rtt_rostopic::ROSTopic rostopic;
my_port_.createStream(rostopic.connection("my_ros_topic"));

and let rostopic query/load the global service itself ?

jbohren: > and let rostopic query/load the global service itself ?

Yes. That's a good point. Fixed in a5389ee

jbohren: So it looks like you can't run `rostopic.connectTo(RTT::internal::GlobalService::Instance()->provides("rostopic"));` in the ServiceRequester's constructor...

jbohren:
Alright, it wasn't a problem with that patch, just another effect. This should be good to merge!

meyerj:
I reviewed these patches and did a few tests with existing packages without any problem. Thanks, @jbohren.

Only one question: Is `control_msgs` part of a "standard" ROS installation (e.g. `ros-hydro-desktop`)? If no, I would prefer to move `rtt_control_msgs` to a separate repository to reduce the number of dependencies. In the commit message you also mentioned `controller_manager_msgs`, but there is no typekit yet for this message package?

jbohren:
> Only one question: Is control_msgs part of a "standard" ROS installation (e.g. `ros-hydro-desktop`)? If no, I would prefer to move `rtt_control_msgs` to a separate repository to reduce the number of dependencies.

I don't think `ros_control` is installed with `ros-hydro-desktop` but it is available as a debian package. We could move it out but since RTT and ros_control are both meant for robot control, I thought it would be good to include these messages as part of the standard distribution. Additionally, while I haven't written it yet, I have plans to write an RTT component that can load ros_control-based controllers. That being said, all of that stuff could be put into another repo for now, called rtt_ros_control or something. Once it becomes useful, then we could merge it back in.

>  In the commit message you also mentioned `controller_manager_msgs`, but there is no typekit yet for this message package?

Yeah that was an oversight. I meant to commit it, but it looks like I didn't. I'll remove rtt_control_msgs, though, and put both typekit packages into a new repo called rtt_ros_control.

jbohren:
I'll re-open this PR once I pull out that package.

meyerj:
I partially merged this pull request in 1f36e062237d9902004ec6150e270d1d471bf5fe and created a new one for the latest commits that add the rtt_rosclock package here: https://github.com/orocos/rtt_ros_integration/pull/20.

PR #20, opened 2014-02-10
Added package rtt_rosclock

meyerj:
This pr is a follow-up of #19.


meyerj:
How does rtt_rosclock reflect the differences between `ros::Time` and `ros::WallTime`? Why is it needed to call `clock_gettime()` directly for Xenomai? In which cases do you expect a difference between `rtt_now()` and `host_rt_now()`?

I think there are three different time sources available:
1. The system time returned by `RTT::TimeService::getNSecs()`. This should be real-time safe for all targets.
2. The ROS wall time returned by `ros::WallTime::now()`. This should always be the same than `RTT::TimeService::getNSecs()` for gnulinux targets as it also calls `clock_gettime(CLOCK_REALTIME)` internally (see https://github.com/ros/roscpp_core/blob/indigo-devel/rostime/src/time.cpp#L96). Might not be real-time safe for other real-time targets.
3. The ROS time returned by `ros::Time::now()`. This is either wall time or simulation time if `/use_sim_time` parameter is set and `ros::Time::isSimTime()` returns true.

In the [rtt_gazebo](https://github.com/jhu-lcsr/rtt_gazebo) package we currently switch the `TimeService` to not use system clock anymore and update it directly from either Gazebo API hooks or from the `/clock` topic. How would that integrate in rtt_rosclock package?

jbohren:
> How does rtt_rosclock reflect the differences between ros::Time and ros::WallTime? Why is it needed to call clock_gettime() directly for Xenomai? In which cases do you expect a difference between rtt_now() and host_rt_now()?

This is all because the clock source used by RTT is the monotonically-increasing real-time clock, which easily diverges from the NTP-synchronized clocks expected in ROS systems. In practice, I've noticed there to be +/- 20 seconds difference between the time reported by RTT and the NTP-synchronized time. 

If you use the RTT time to broadcast something like a TF frame or any other temporal information that you expect to be used by other machines, you'll end up with undefined behavior like large delays in when something has happened or when it's supposed to happen.

In Xenomai 2.6 they added the ability to safely query the NTP-synchronized time with the `CLOCK_HOST_REALTIME` clock source. This allows xenomai-based RTT tasks to send information out to non-realtime ROS nodes and ROS nodes on other computers which are synchronized via NTP.

> I think there are three different time sources available:

Right, and we can rearrange this list by their origins.

When build against `gnulinux` we have only 2 time sources:
1. The **real-time clock time** returned by `clock_gettime(CLOCK_REALTIME)` ( `RTT::TimeService::getNSecs()`, `ros::WallTime::now()`, and `ros::Time::now()` when not in simulation.
2. The **ROS simulation time** from the `/clock` topic when in simulation.

When built against `xenomai` we have _3_ time sources:
1. The **real-time clock time** returned by `clock_gettime(CLOCK_REALTIME)` ( `RTT::TimeService::getNSecs()`, `ros::WallTime::now()`, and `ros::Time::now()` when not in simulation.
2. The **NTP-synchronized real-time-safe time** returned by `clock_gettime(CLOCK_HOST_REALTIME)`.
3. The **ROS simulation time** (`/clock` topic) when in simulation.

I've been running this for a while now for testing, and I'd like to integrate it with the time service in [rtt_gazebo](https://github.com/jhu-lcsr/rtt_gazebo), but I just haven't gotten around to it yet.

jbohren:
What I'm wondering now is how `ros::WallTime` calling `clock_gettime(CLOCK_REALTIME)` when it's linked against the xenomai libraries with the posix skin. I need to run a quick test to see if that's actually the real-time clock time if it's called from a xenomai task.

meyerj:
Okay, I just wondered why `host_rt_now()` returns `ros::Time::now()` in the non-Xenomai case instead of `ros::WallTime::now()`.

jbohren:
> Okay, I just wondered why host_rt_now() returns ros::Time::now() in the non-Xenomai case instead of ros::WallTime::now().

Ah yeah, I guess to be consistent that should return ros::WallTime::now(). The only reason I'm not making that distinction is because I wasn't trying to handle how it would work in simulation yet.

meyerj:
I guess this pr can wait until after the first release of rtt_ros_integration in hydro (finally)?

One issue that I just noticed in rtt_ros is that `ros.time.now()` does not work from scripting (at least not in the deployer):

```
Deployer [S]> ros.time.now()
The string "time" cannot be used as an identifer.
```

I already moved the primitive typekit from rtt_roscomm to rtt_ros in a541d0092a0825fcc6bb49702cb8918dcdd7abb1 to be able to use the `time` and `duration` types without rtt_roscomm.

psoetens:
'time' is a reserved keyword in classical scripting. We could remove it from that list easily without impairing / changing the behavior of scripting. It's only used in 'until' clauses for limiting asynchronous waits on operation completions. Removing time from the keyword list would ignore a 'time' variable or function in this until clause and use it as a keyword anyway.

 It has bitten other users as well...

jbohren:
> 'time' is a reserved keyword in classical scripting

Yeah, this is why I started using the "clock" designation for these tools instead.

konradb3:
Is it this PR moving forward ? I (and probably everyone using RTT, Xenomai and ROS ) would like to see this or something similar merged soon, so i could migrate my packages to more standardized way of handling time in Xenomai.

jbohren:
Yeah it is. After we resolve #21, I'm going to combine this with the time tools from rtt_gazebo. Even before that, these functions will be merged in as part of #21.

jbohren:
We can close this in favor of #26

PR #21, opened 2014-02-10
Collect all global ROS operations in service ros

meyerj:
The hydro-devel branch introduced the new `rostopic` global service that provides operations to create ROS streaming ConnPolicies. I would prefer to use the global service `ros` instead to collect all kind of ROS-related operations that not require to be loaded in a specific component. With this patch the rostopic plugin only installs operations in the global service "ros" instead of adding a new global service. The packages can (and must) still be imported independently:

```
import("rtt_ros") # installs global service "ros" and the ros.import() operation
ros.import("rtt_std_msgs") # imports rtt_roscomm package from rtt_std_msgs's <plugin_depend>
                           # and adds operations ros.topic, ros.topicBuffer and ros.topicUnbuffered

stream("foo.bar", ros.topic("bar"))
...
```

The main driver for this proposal is backwards compatibility with scripts generated for older versions of the toolchain. And I personally liked the `ros.topic(...)` syntax more than `rostopic.connection(...)`.

The ServiceRequester `rtt_rostopic::ROSTopic` would still work and uses the new names for the OperationCallers.

Just a suggestion. What do you think?

### DIFF
--- a/rtt_ros/src/rtt_ros_service.cpp
+++ b/rtt_ros/src/rtt_ros_service.cpp
@@ -12,15 +12,10 @@
 #include <libxml/tree.h>
 
 #include <rtt/RTT.hpp>
-#include <rtt/plugin/ServicePlugin.hpp>
 #include <rtt/internal/GlobalService.hpp>
 
 #include <rtt/deployment/ComponentLoader.hpp>
-#include <rtt/TaskContext.hpp>
 #include <rtt/Logger.hpp>
-#include <rtt/os/StartStopManager.hpp>
-#include <rtt/plugin/PluginLoader.hpp>
-#include <rtt/types/TypekitRepository.hpp>
 
 #include <rospack/rospack.h>
 
@@ -32,32 +27,13 @@ using namespace std;
 /**
  * The globally loadable ROS service.
  */
-class ROSService : public RTT::Service {
+class ROSService {
 public:
-  int protocol_id;
-  /**
-   * Instantiates this service.
-   * @param owner The owner or null in case of global.
-   */
-  ROSService(TaskContext* owner) 
-    : Service("ros", owner)
-  {
-    this->doc("RTT service for loading RTT plugins ");
-
-    // ROS Package-importing
-    this->addOperation("import", &ROSService::import, this).doc(
-        "Imports the Orocos plugins from a given ROS package (if found) along with the plugins of all of the package's run or exec dependencies as listed in the package.xml.").arg(
-            "package", "The ROS package name.");
-
-    this->provides("time")->addOperation("now", &rtt_ros::time::now).doc(
-        "Get a ros::Time structure based on the RTT time source.");
-  }
-
   /**
    * Returns a ConnPolicy object for streaming to or from 
    * the given ROS topic. No buffering is done.
    */
-  bool import(const std::string& package) 
+  static bool import(const std::string& package)
   {
     RTT::Logger::In in("ROSService::import(\""+package+"\")");
 
@@ -235,13 +211,23 @@ public:
 };
 
 void loadROSService(){
-  RTT::Service::shared_ptr rts(new ROSService(0));
-  RTT::internal::GlobalService::Instance()->addService(rts);
+  RTT::Service::shared_ptr ros = RTT::internal::GlobalService::Instance()->provides("ros");
+
+  ros->doc("RTT service for loading RTT plugins ");
+
+  // ROS Package-importing
+  ros->addOperation("import", &ROSService::import).doc(
+      "Imports the Orocos plugins from a given ROS package (if found) along with the plugins of all of the package's run or exec dependencies as listed in the package.xml.").arg(
+          "package", "The ROS package name.");
+
+  ros->provides("time")->addOperation("now", &rtt_ros::time::now).doc(
+      "Get a ros::Time structure based on the RTT time source.");
 }
 
 using namespace RTT;
 extern "C" {
   bool loadRTTPlugin(RTT::TaskContext* c){
+    if (c != 0) return false;
     loadROSService();
     return true;
   }
--- a/rtt_roscomm/include/rtt_rostopic/rostopic.h
+++ b/rtt_roscomm/include/rtt_rostopic/rostopic.h
@@ -11,18 +11,18 @@ namespace rtt_rostopic {
   class ROSTopic : public RTT::ServiceRequester
   {
   public:
-    ROSTopic() :
-      RTT::ServiceRequester("rostopic",NULL),
-      connection("connection"),
-      bufferedConnection("bufferedConnection"),
-      unbufferedConnection("unbufferedConnection"),
+    ROSTopic(RTT::TaskContext *owner = 0) :
+      RTT::ServiceRequester("rostopic", owner),
+      connection("topic"),
+      bufferedConnection("topicBuffer"),
+      unbufferedConnection("topicUnbuffered"),
       protocol_id(ORO_ROS_PROTOCOL_ID)
     {
       this->addOperationCaller(connection);
       this->addOperationCaller(bufferedConnection);
       this->addOperationCaller(unbufferedConnection);
 
-      this->connectTo(RTT::internal::GlobalService::Instance()->provides("rostopic"));
+      this->connectTo(RTT::internal::GlobalService::Instance()->provides("ros"));
     }
 
     RTT::OperationCaller<RTT::ConnPolicy(const std::string &)> connection;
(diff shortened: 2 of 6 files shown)


jbohren:
So I like the idea of merging all of the core functionality into a single package / plugin. I also like that this resolves some of the backwards-compatibility issues created by the re-factor. 

With respect to the ROS topic connections, I'm fine with bringing back the old `ros.topic(...)` API, but I would still like to have an explicit `ros.topicConnection(...)` or `ros.topicConnPolicy()` since that  operation is actually returning a `ConnPolicy`.

jbohren:
So if these services all add their operations to the `ros` global service, then should we combine them into a single package?

meyerj:
I also thought about that, but on the other side the global `ros` service is just a collection of ROS related operations and multiple plugin libraries in different packages can contribute operations. This is how I implemented it now in [rtt_ros_service.cpp:213](https://github.com/meyerj/rtt_ros_integration/blob/all-operations-in-global-ros-service/rtt_ros/src/rtt_ros_service.cpp#L213) and [rtt_rostopic_service.cpp:54](https://github.com/meyerj/rtt_ros_integration/blob/all-operations-in-global-ros-service/rtt_roscomm/src/rtt_rostopic_service.cpp#L54). You still have to import the individual packages separately or through plugin dependencies using the `ros.import()` operation. [rtt_rosclock](https://github.com/orocos/rtt_ros_integration/tree/rtt-rosclock/rtt_rosclock) would be another candidate to add clock-related operations instead of installing its own global service. This works for all cases where the operations can be implemented using static functions. Any objectives?

So, I would suggest to remove the global `rostopic` service, bring back
- `ros.topic`
- `ros.topicBuffered` and
- `ros.topicUnbuffered`.

In addition to the current pull request we could add aliases (operations that point to the same implementation):
- `ros.topicConnection`
- `ros.topicBufferedConnection` and
- `ros.topicUnbufferedConnection`.

The only question is if the OperationCallers in the `ROSTopic` service requester should also be renamed accordingly, like

``` cpp
#include <rtt_rostopic/rostopic.h>
ROSTopic rostopic;
rostopic.topicConnection(...);            // was: rostopic.connection(...)
rostopic.topicBufferedConnection(...);    // was: rostopic.bufferedConnection(...)
rostopic.topicUnbufferedConnection(...);  // was: rostopic.unbufferedConnection(...)
```

jbohren:
@meyerj 

> You still have to import the individual packages separately or through plugin dependencies using the ros.import() operation. rtt_rosclock would be another candidate to add clock-related operations instead of installing its own global service. This works for all cases where the operations can be implemented using static functions.

Well, for the sake of transparency, I feel like if functionality is added to the `ros` global service then that functionality should be implemented in the same (single) package, even if they're separate plugins in that package. So `rtt_ros`, `rtt_rostopic`, `rtt_rosclock` could remain separate plugins, but they all reside in the `rtt_ros` package. 

It means things stay modular, but it's less likely that people get confused about which plugins are loaded by which package. Additionally, it means that `rtt_ros` can expose a _single_ service requester which incorporates all of these plugins, so that the C++ interface matches the dynamic orocos one.

smits:
I agree with Johannes, I don't see a technical reason why all services should be part of one package. From a package dependency point of view I would even object against it.

meyerj:
> Additionally, it means that rtt_ros can expose a single service requester which incorporates all of these plugins, so that the C++ interface matches the dynamic orocos one.

@jbohren: It is no problem that a service requester is connected to a service that provides more than the required operations. The [ROSTopic](https://github.com/orocos/rtt_ros_integration/pull/21/files#diff-43ae062646292bf6fda23f81527a8e5a) service requester works like before. It just has to connect to a different service.

> From a package dependency point of view I would even object against it.

rtt_ros should only depend on a minimal set of packages including rtt and ocl. Especially it  should not depend on roscpp. For `rtt_rostopic` and `rtt_rosclock` we would not need new requirements as far as I see, so it would also be fine for me to integrate these plugins in the `rtt_ros` package, but there is no technical reason why we have to do that.

We could also add the services as a sub-service in the global `ros` service, so that you can invoke operations like `ros.clock.now()`?

jbohren:
I like the idea of gathering things under subservices, and if you guys want to keep the packages lightweight, that's still fine with me. On that case, however, if the currently-named 'rosxxx.yyy' services are going to be accessed as 'ros.xxx.yyy' then I propose that the packages themselves are called 'rtt_ros_xxx' instead of 'rtt_rosxxx'. So 'rtt_rosclock' would be 'rtt_ros_clock'. How does that sound?

meyerj:
> On that case, however, if the currently-named 'rosxxx.yyy' services are going to be accessed as 'ros.xxx.yyy' ...

With the exemption of the `ros.topic()` operation for sake of backwards compatibility. I am not sure what would be the result of `ros.topic` in scripting if there is also a service named `ros.topic`.

> ... then I propose that the packages themselves are called 'rtt_ros_xxx' instead of 'rtt_rosxxx'. So 'rtt_rosclock' would be 'rtt_ros_clock'. How does that sound?

I would not rename packages at this point anymore as this requires a rewrite of numerous package manifests. The standard dependencies `rtt_rosnode` and `rtt_*_msgs` in older releases are still fine if only the deployment scripts are updated to use `ros.import(...)`.

By the way, as most operations currently provided in rtt_ros_integration are just global function calls at the end, do we really need to collect them in services and write service requesters to make them easily accessible from C++? What is the advantage compared to just provide a header to call them directly in C++ and add an operation to the `ros` global service for scripting? For rtt_ros and rtt_rostopic this could be as simple as

``` cpp
#include <rtt_ros/import.h>
#include <rtt_rostopic/rostopic.h>
rtt_ros::import("rtt_std_msgs");
ConnPolicy cp = rtt_rostopic::connection(...);
```

or alternatively

``` cpp
#include <rtt_ros/import.h>
#include <rtt_ros/topic.h>   // <-- all headers are installed in include/rtt_ros/ independent of the package name
rtt_ros::import("rtt_std_msgs");
ConnPolicy cp = rtt_ros::topic::connection(...); // <-- nested namespaces
```

where scripting uses

```
ros.import("rtt_std_msgs")
var ConnPolicy cp = ros.topic.connection(...)
```

jbohren:
@meyerj 

> I would not rename packages at this point anymore as this requires a rewrite of numerous package manifests. The standard dependencies rtt_rosnode and rtt_*_msgs in older releases are still fine if only the deployment scripts are updated to use ros.import(...).

Yeah, that's reasonable. So then the idea is for resources in a given package (where in some cases `xxx` and `yyy` might be identical):
- package named: `rtt_rosxxx`
- header for global variables/functions: `rtt_rosxxx/rtt_rosyyy.h`
- namespace for global variables/functions: `rtt_rosxxx`
- global variables/functions access from orocos script: `ros.xxx.*`
- header for task-bound service requester: `rtt_rosxxx/rosyyy.h`
- task-bound variables/functions access from orocos script: `my_task.rosyyy.*`
- service for task-bound services: `rtt_rosxxx::ROSYyyService`
- service requester named: `rtt_rosxxx::ROSYyy`

This will handle the backwards-compatibility issues with things like `ros.topic` since there is no `rtt_rostopic` package. The new operations could be under `ros.comm.*` for creating different types of connection policies, and `ros.topic` can hang around for backwards-compatibility.

I know things shouldn't change a lot, but I regret creating include directories in `rtt_roscomm` not named `rtt_roscomm`. As such, the headers from the `rtt_roscomm` pacakge look like this:

```
.
├── rtt_rosservice
│   ├── rosservice.h
│   ├── rtt_rosservice_proxy.h
│   └── rtt_rosservice_registry_service.h
└── rtt_rostopic
    ├── ros_msg_transporter.hpp
    ├── ros_publish_activity.hpp
    ├── rostopic.h
    └── rtt_rostopic.h
```

I propose that as part of this reorganization, we combine these two directories into a single `rtt_roscomm` header directory. I don't think this will cause too many issues, since typekit code is now auto-generated at build time.

@meyerj 

> By the way, as most operations currently provided in rtt_ros_integration are just global function calls at the end, do we really need to collect them in services and write service requesters to make them easily accessible from C++? What is the advantage compared to just provide a header to call them directly in C++ and add an operation to the `ros` global service for scripting? For rtt_ros and rtt_rostopic this could be as simple as

Yeah, this is a really good point. I'd be happy dropping all the global service requesters in favor of static functions. I'm at a point where I can put some time back into this infrastructure tomorrow (Feb 26). I'm going to prototype these changes in a fork of your `all-operations-in-global-ros-service` fork.

meyerj:
Superseded by #26.

PR #26, opened 2014-02-26
Global service refactor

jbohren:
This PR includes changes proposed in #21 as well as the rtt_rosclock from #20.

It refactors the rtt_roscomm package and makes all global services put their operations under the `ros` service. Instead of service requesters, their static functions are exposed directly in header files.


rtt_ros/CMakeLists.txt
```diff
@@ -1,14 +1,13 @@
...
 
-find_package(catkin REQUIRED COMPONENTS rospack rostime)
+find_package(catkin REQUIRED COMPONENTS rospack rostime cmake_modules roslib)
 
 find_package(OROCOS-RTT REQUIRED 
   COMPONENTS rtt-marshalling rtt-scripting)
 include(${OROCOS-RTT_USE_FILE_PATH}/UseOROCOS-RTT.cmake)
 
 catkin_package(
-  INCLUDE_DIRS include
```
meyerj, line 11: Why did you remove this line? The `rtt_ros/rtt_ros.h` header should be exported to using packages?

meyerj:
Thanks, @jbohren, for having done this refactoring.

I did not really understand the intention behind https://github.com/jhu-lcsr-forks/rtt_ros_integration/commit/8d32ba953e2a9253b663afaf150e3f4320624f91.

Why did you move the implementation of `rtt_ros::import()` into the header? Packages that depend on rtt_ros will link to the `librtt_ros-gnulinux.so` library automatically if they use `rtt_ros`, even if it is a service library.

jbohren:
Ah, the message was about moving the declaration into the header, there's no reason why the  implementation can't go into the rtt_ros.cpp source file.

PR #27, opened 2014-02-27
rtt_rosclock: Adding SimClockActivity

jbohren:
This PR expands rtt_rosclock to incorporate a fork of @meyerj's rtt_gazebo GazeboActivity with some modifications that enable it to use either the ROS `/clock` topic or a manual clock source if there is one in-process. See the [rtt_rosclock/README.md](https://github.com/jhu-lcsr-forks/rtt_ros_integration/blob/gsr-sim-clock/rtt_rosclock/README.md) for example usage.

This PR also incorporates changes included in https://github.com/orocos/rtt_ros_integration/pull/26


smits:
Looks ok to me, but I'll let @meyerj decide on and do the merge ;)

jbohren:
Cool. I'm still testing it out, I was worried about doing all the time updates for the simulation clock with doubles, and I've found some cases where there are floating point precision errors.

jbohren:
So if I call `TimeService::enableSystemClock(false)` too early, it doesn't actually take effect. Does that sound familiar?

jbohren:
It looks like the issue is with calling `secondsChange()` and not actually changing the clock precisely. I've opened an MR over on gitorious which adds a similar `ticksChange()` method to `RTT::os::TimeService` as well as an accessor to check if the system clock is being used: https://gitorious.org/orocos-toolchain/rtt/merge_requests/27

meyerj:
I merged this pull request with the latest version of #26 with my additional patches in https://github.com/jhu-lcsr-forks/rtt_ros_integration/pull/58 and the latest commits in the [hydro-devel](https://github.com/orocos/rtt_ros_integration/compare/jhu-lcsr-forks:gsr-sim-clock...hydro-devel) branch:
- https://github.com/meyerj/rtt_ros_integration/tree/gsr-sim-clock
- Comparison with hydro-devel: https://github.com/meyerj/rtt_ros_integration/compare/orocos:hydro-devel...gsr-sim-clock

The new `SimClockActivity` works fine in my current application where I used `rtt_gazebo_activity` before.

jbohren:
> The new `SimClockActivity` works fine in my current application where I used `rtt_gazebo_activity` before.

Cool! I've been testing it here, too, running in-process: https://github.com/jhu-lcsr/orocos_barrett/tree/master/oro_barrett_sim

PR #29, opened 2014-03-31
Numerous fixes to rtt_tf's tf_interface, also filling out rtt_rosparam's service requester

jbohren:
- **rtt_tf:** Adding `canTransform` RTT_TF operation
- **rtt_rosparam:** Adding `getParam` and `setParam` for getting and setting params which have different names (or namespaces) from their property names
- **rtt_rosdeployment:** Adding a plugin which lets you manipulate an OCC::DeploymentComponent over ROS service calls (like run scripts, list components, etc)
- **rtt_ros_msgs:** .srv types used by rtt_rosdeployment
- **rtt_rosclock:** **_significant change see new rtt_rosclock README.md**_ clarifying which operations handle wall time vs sim time

### DIFF
--- a/rtt_rosclock/src/rtt_rosclock.cpp
+++ b/rtt_rosclock/src/rtt_rosclock.cpp
@@ -51,10 +51,12 @@ const ros::Time rtt_rosclock::host_wall_now()
 
 const ros::Time rtt_rosclock::rtt_now() 
 {
+  // count the zeros...   -987654321--
+  const uint64_t one_E9 = 1000000000ll;
   // NOTE: getNSecs returns wall time, getTicks returns offset time
   uint64_t nsec64 = RTT::os::TimeService::ticks2nsecs(RTT::os::TimeService::Instance()->getTicks()); //RTT::os::TimeService::Instance()->getNSecs();
-  uint32_t sec = nsec64 / 1E9;
-  uint32_t nsec = (uint32_t)(nsec64 - (sec*1E9));
+  uint32_t sec = nsec64 / one_E9;
+  uint32_t nsec = (uint32_t)(nsec64 - (sec*one_E9));
   return ros::Time(sec, nsec);
 }
 
--- a/rtt_rosclock/src/rtt_rosclock.cpp
+++ b/rtt_rosclock/src/rtt_rosclock.cpp
@@ -10,44 +10,66 @@ namespace rtt_rosclock {
   boost::shared_ptr<rtt_rosclock::SimClockThread> sim_clock_thread;
 }
 
+const ros::Time rtt_rosclock::host_now()
+{
+  if(SimClockThread::GetInstance() && SimClockThread::GetInstance()->simTimeEnabled()) {
+    return rtt_now();   
+  }
+
+  #ifdef __XENO__
+    // Use Xenomai 2.6 feature to get the NTP-synched real-time clock
+    timespec ts = {0,0};
+    int ret = clock_gettime(CLOCK_HOST_REALTIME, &ts);
+    if(ret) {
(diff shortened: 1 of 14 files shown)


jbohren:
I've added a bunch of things into this, see the updated description and package README files for more details.

jbohren:
Bump

rtt_rosclock/src/rtt_rosclock.cpp
```diff
@@ -10,44 +10,66 @@ namespace rtt_rosclock {
...
     return ros::Time(now.sec, now.nsec);
-    #endif
-  }
+  #endif
+}
+
+const ros::Time rtt_rosclock::rtt_now() 
+{
+  // NOTE: getNSecs returns wall time, getTicks returns offset time
+  uint64_t nsec64 = RTT::os::TimeService::ticks2nsecs(RTT::os::TimeService::Instance()->getTicks()); //RTT::os::TimeService::Instance()->getNSecs();
```
meyerj, line 57: In `rtt_rosclock::rtt_now()` and `rtt_rosclock::rtt_wall_now()`:

I would suggest to use type `RTT::nsecs` here instead of `uint64_t`, which is the return type of `ticks2nsecs()` and resolves to `long long`. Furthermore `1E9` should be replaced by `1000000000ll` as constants in scientific notation are floating point constants, making the operations floating point instead of integer operations.

rtt_rosdeployment/src/rtt_rosdeployment_service.cpp
```diff
@@ -0,0 +1,138 @@
...
+*      this software without specific prior written permission.
+*     THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING,
+*     BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE 
+*     ARE DISCLAIMED. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+*     EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+*     OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY 
+*     THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR 
+*     OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY 
+*     OF SUCH DAMAGE.
+**/
```
meyerj: This license header looks like copied from somewhere else? According to the `package.xml` file `rtt_rosdeployment` has a BSD license?

rtt_rosclock/src/rtt_rosclock.cpp
jbohren, line 57: All sounds reasonable.

rtt_rosdeployment/src/rtt_rosdeployment_service.cpp
jbohren: I'll resolve that.

rtt_rosclock/src/rtt_rosclock.cpp
jbohren, line 57: So upon review of this, I think I want to stick with uint64_t since it's meant to ensure that the separation below into sec and nsec happens correctly with the uint32_t. I am, however, changing `1E9` to `1000000000ll`.

jbohren:
Resolved @meyerj's concerns. Anything else before this gets merged?

PR #30, opened 2014-05-08
Update README.md

b1willaert:


b1willaert:
Error in documentation line: create_rtt_pkg instead of create_rtt_msg

smits:
I would even vote for renaming create_rtt_pkg to create_rtt_typekit.

smits:
Hmm, I guess you are not using the latest version of rtt_ros_integration, as the script is currently called create_rtt_msgs, see 695f1e2

PR #37, opened 2014-10-09
Numerous updates + "ldeployer"

jbohren:
- Adding missing target dep on rtt_rosclock to rtt_actionlib
- Adding names to action server threads
- Fixing a problem with ticks2nsecs on recent versions of Xenomai

_removed ldeployer in favor of lua suport for deployer_

> -I've also added a lua deployer called `ldeployer`. This is like the normal `deployer` program, except it runs lua scripts. This is different than `rttlua` because it provides a `TaskBrowser` instead of a Lua command-line interface. It's equivalent to running the `deployer`, loading the `Lua` service on the `Deployer` component, and then running the scripts. I put this in here for now, but it might make more sense to just add an additional command-line option to the normal `deployer` to run Lua files. I can also split off the `ldeployer` changes from this PR if desired.


meyerj:
Hi Jonathan, I must have overlooked this pull request. All changes seam reasonable and useful to me.

Is there a reason not to add the new `ldeployer` to ocl directly and only add the usual startup script to rtt_ros as for `deployer`, `rttlua` and `rttscript`? If possible, I personally would prefer your suggestion to add Lua as a new script file type to the standard deployer as it already supports xml and ops deployments.

We are going to merge and finally release Orocos Toolchain 2.8 and rtt_ros_integration in indigo tomorrow.

meyerj:
I merged the first three commits of this PR to hydro-devel. As suggested, I also added Lua support to the standard deployer (see pull request https://github.com/orocos-toolchain/ocl/issues/27).

@jbohren, could you please review https://github.com/orocos-toolchain/ocl/issues/27 and check if this would be an acceptable solution for you? The standard deployer does not support the `--define-globals` command line argument. I don't think it is required as the deployer supports multiple script files and the global variables could be easily defined by an additional script loaded before all the others.

The only thing missing from this PR would be the the new launch files added in e504a7f7b1156beacd405d345bf78956aa94003a then. Could you please rebase it or create a new pull request?

jbohren:
This is just the additional scripts and launchfiles now.

meyerj:
Merged without the bad merge in rtt_actionlib.
- hydro-devel: https://github.com/orocos/rtt_ros_integration/commit/b187ce6827de6ebe779ee9222d6482b189c3406d
- indigo-devel: https://github.com/orocos/rtt_ros_integration/commit/225370a03413f5f1a9961d60cb092fff4e30a2b2

jbohren:
Rad.

PR #39, opened 2015-05-17
rtt_dynamic_reconfigure: added support for Property composition and decomposition and fixed reconfiguration of nested properties

meyerj:
Reconfiguration of nested properties (properties inside a PropertyBag) was broken because `AutoConfig::__fromMessage__()` aborted too early if a property name was not found in the Config message, even if it is the name of a nested bag.

Composition support now allows to use the reconfigure service even for properties with complex types that support composition (e.g. types from kdl_typekit or eigen_typekit).

This patch also applies to indigo-devel. I will merge it as soon as it has been accepted for hydro.

### DIFF
--- a/rtt_dynamic_reconfigure/include/orocos/rtt_dynamic_reconfigure/auto_config.h
+++ b/rtt_dynamic_reconfigure/include/orocos/rtt_dynamic_reconfigure/auto_config.h
@@ -76,6 +76,9 @@ public:
 
     static void __refreshDescription__(const ServerType *server);
 
+    bool updateProperties(RTT::PropertyBag &) const;
+    bool fromProperties(const RTT::PropertyBag &);
+
 private:
     struct Cache;
     typedef boost::shared_ptr<Cache> CachePtr;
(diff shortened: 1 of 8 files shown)


meyerj:
This patch breaks reconfiguration of bool properties and the unit test in rtt_dynamic_reconfigure_tests will fail until https://github.com/orocos-toolchain/rtt/pull/95 is merged.

SvenDC:
When adding properties to my_component AFTER having loaded the rtt_dynamic_reconfigure service in my_component, rtt_dynamic_reconfigure does not pick up these new properties.
This is workable but not truly dynamic ;)

meyerj:
The "dynamic" in dynamic_reconfigure stands for dynamic upgrades without restarting the whole application, not for adding and removing parameters during runtime. However, rtt_dynamic_reconfigure allows this (not necessarily the user interface), but not automatically.

Did you try to call the `reconfigure.refresh()` operation after having added or removed properties?

SvenDC:
Yes you are right calling refresh() works. Thanks for the reply.

PR #41, opened 2015-06-01
fix rtt_now when sim clock is running on xenomai

ahoarau:
Discussed here : 
https://github.com/jhu-lcsr/rtt_gazebo/issues/21

### DIFF
--- a/rtt_rosclock/src/rtt_rosclock.cpp
+++ b/rtt_rosclock/src/rtt_rosclock.cpp
@@ -55,13 +55,17 @@ const ros::Time rtt_rosclock::rtt_now()
   const uint64_t one_E9 = 1000000000ULL;
   // NOTE: getNSecs returns wall time, getTicks returns offset time
 #ifdef __XENO__
-  uint64_t nsec64 = RTT::os::TimeService::Instance()->getNSecs(); //RTT::os::TimeService::Instance()->getNSecs();
+  uint64_t nsec64;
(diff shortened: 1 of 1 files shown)


rtt_rosclock/src/rtt_rosclock.cpp
```diff
@@ -55,13 +55,17 @@ const ros::Time rtt_rosclock::rtt_now()
   const uint64_t one_E9 = 1000000000ULL;
   // NOTE: getNSecs returns wall time, getTicks returns offset time
 #ifdef __XENO__
```
meyerj, line 57: Why is this distinction between Xenomai and non-Xenomai builds required here at all, independent of sim time?
Wouldn't `getTicks()` (the else branch) work for both cases?

`getTicks()` => `rt_timer_tsc()` (if system clock is enabled in the TimeService)
`getNSecs()` => `rt_timer_ticks2ns(rt_timer_read())` (independent of sim/system time mode)

ahoarau, line 57: I'll look into that next week. Thanks for the clarification !

ahoarau, line 57: That's very true, I'm closing this PR until I provide a new one with your recommendations :)

ahoarau:
fixed PR with https://github.com/orocos/rtt_ros_integration/pull/65

PR #60, opened 2016-05-31
Added explicit destructors to fix #59

### LINKED ISSUE #59: Cleaning up registered services
I am interested in dynamically loading and unloading components.
This works nice for standard components with `d:loadComponent()` and `d:unloadComponent()`, but when a `ROSServiceService` was loaded into a component, the service proxy seems to not "unadvertise" correctly on component unload and throws errors if the same service "topic" is used.

For instance, I added some controller manager orocos_services to a component by following this example https://github.com/jbohren/conman/blob/master/conman_ros/src/ros_interface_service.cpp  
This orocos_service uses the `ROSServiceService` to connect, meaning that it advertises a ros service through a `NodeHandle`.

However on unload, there is no disconnect in `ROSServiceService` and the `NodeHandle` apparently does not shutdown correctly so that I get errors if I redeploy the same component+orocos_service.

```
[ERROR] [1464168075.512551730, 1005.620000000]: Tried to advertise a service that is already advertised in this node [/ra/controller_manager/list_controller_types]
```

even if the component does not appear anymore (neither in `rttlib.stat()`  nor in `d:getPeers()`)

You might argue that a controller manager should always be there even in the case of dynamic deployment (conman does that by managing the list of peers and such), but the problem goes beyond the functionality of a controller manager, any orocos_service could create a `ROSServiceService` that at some point might needs to be cleaned-up for redeployment later on when using dynamic deployment. 

What would be a correct way to unadvertise the service ? at which level should it happen ? I could then try to modify the `rtt_roscomm` to handle that and create a PR.

guihomework:
Following suggestions in #59 , I added two explicit destructors. Adding only the one in the proxy did not suffice as the pointers stored in the map were not explicitly deleted. 

Changing to shared_ptr seemed like a risky think to do for me, and would cause API change as far as I understand. Not an expert, so I am not usre if my change is an ABI change, but at least it is not an API change. 

If everything is ok for you, then I am interested to have this fix back-ported to indigo if possible.

### DIFF
--- a/rtt_roscomm/include/rtt_roscomm/rtt_rosservice_proxy.h
+++ b/rtt_roscomm/include/rtt_roscomm/rtt_rosservice_proxy.h
@@ -11,6 +11,7 @@ class ROSServiceProxyBase
 {
 public:
   ROSServiceProxyBase(const std::string &service_name) : service_name_(service_name) { }
+  virtual ~ROSServiceProxyBase() { }
   //! Get the name of the ROS service
   const std::string& getServiceName() const { return service_name_; }
 private:
--- a/rtt_roscomm/include/rtt_roscomm/rtt_rosservice_proxy.h
+++ b/rtt_roscomm/include/rtt_roscomm/rtt_rosservice_proxy.h
@@ -67,6 +67,12 @@ public:
         this);
   }
 
+  ~ROSServiceServerProxy()
+  {
+    // Clean-up advertized ROS services
+    server_.shutdown();     
+  }
+
 private:
   
   //! The callback called by the ROS service server when this service is invoked
--- a/rtt_roscomm/src/rtt_rosservice_service.cpp
+++ b/rtt_roscomm/src/rtt_rosservice_service.cpp
@@ -38,6 +38,22 @@ public:
     get_service_factory = rosservice_registry_->getOperation("getServiceFactory");
   }
 
+  ~ROSServiceService()
+  {
+    // Cleanup registered ROS services and clients
+    std::map<std::string, ROSServiceServerProxyBase*>::iterator iter_s;
+    for(iter_s=server_proxies_.begin(); iter_s != server_proxies_.end(); iter_s++)
+    {
+      delete iter_s->second;
+    }
+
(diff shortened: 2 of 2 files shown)


meyerj:
The destructor of `ROSServiceServerProxy` is indeed also required because the ROS ServiceServer lives in the base class `ROSServiceServerProxyBase` and there is a small chance that a spinner calls the `ros_service_callback()` member function on an instance that is already destroyed. What is missing is a virtual destructor (with an empty body) in the base class `ROSServiceServerProxyBase` or even better `ROSServiceProxyBase`. Otherwise the child's destructor will not be called if a pointer to the base class is used to delete the instances in `~ROSServiceService()`.

Could you please add this virtual destructor to `ROSServiceProxyBase`? Otherwise I will do it tomorrow and merge this PR to indigo-devel and jade-devel (they are identical).

With the virtual destructors the patch breaks ABI of the proxy classes. But at least in the most common case of a service plugin there is no problem because they are only used internally by the factories and in the same compilation unit. In cases where the `rtt_roscomm/rtt_rosservice_proxy.h` header has been included and used from other packages, they would have to be recompiled.

guihomework:
Will do that now

guihomework:
I am not sure why the ROSServiceServerProxyBase does not need a virtual destructor either. After all this is the one that gets deleted when iterating through the map. I can add it if you confirm it is needed.

meyerj:
It is sufficient that the base class `ROSServiceProxyBase` has a virtual destructor (e.g. from http://stackoverflow.com/questions/2198379/are-virtual-destructors-inherited) and `ROSServiceServerProxyBase` and `ROSServiceClientProxyBase` will have an automatically generated virtual destructor, too. Of course it would not hurt to define them explicitly, but it is just simplest to declare the virtual constructor exactly once at the root of the inheritance tree. This is especially important if you cannot exclude that a user of the class ever deletes an instance with a `ROSServiceProxyBase` pointer.

guihomework:
thanks for the explanation and pointers to good resources. Learning while fixing is nice :-)

PR #65, opened 2016-07-08
fix xenomai rtt_now() fix PR#41

ahoarau:
I took care of your remarks regarding the getTicks function. 
It indeed works for both cases.


meyerj:
lgtm, thanks.

PR #69, opened 2016-07-20
rtt_roscomm: set minimum ROS subscriber queue_size to 1

meyerj:
Using a default queue size of 0 is dangerous. See http://wiki.ros.org/rospy/Overview/Publishers%20and%20Subscribers#Choosing_a_good_queue_size.

I would like to release this patch together with #67 in ROS indigo, knowing that is has the potential to break existing Orocos applications that somehow rely on roscpp's internal buffering, but actually the behavior is undefined anyway if the RTT internally uses data connections and drops older samples after the ROS spinner passes them from roscpp to RTT.


meyerj:
Briefly discussed this with @smits and he agrees with the patch.

meyerj:
Actually the branch should have been named `fix-rtt_roscomm-queue-size`.

PR #74, opened 2016-08-16
rtt_roscomm: CORBA transport patches

meyerj:
Follow-up of #51:

The code for conversion from ROS to CORBA OctetSeq was overly complex. Patch https://github.com/orocos/rtt_ros_integration/commit/82036af389b028187dcbe1ae45c77626d1902e30 avoids one copy of the sequence and directly serializes the message into the memory reserved by the CORBA sequence type.

https://github.com/orocos/rtt_ros_integration/commit/a1ac3717a2683976a821854a6f5756d5896ca5b2 removes the need to generate headers and source code from IDL. The CORBA transport of ROS messages serializes all messages into a `CORBA::OctetSeq`, so there is no need for generated code.
@ahoarau: What was the reason to introduce a custom IDL file?

The other two minor commits simplify code reuse if the templates or generated headers are included from another package.


ahoarau:
I initially started without these files ... 

I posted a thread a while ago http://www.orocos.org/forum/rtt/rtt-dev/rttgeometry-idl-files-necessity
I admit that i did not fully understand the answer :^o

PR #75, opened 2016-08-18
RosSubChannelElement should be created as a shared_ptr

### LINKED ISSUE #61: rtt_roscomm cleaning up registered topics
Similar to #59 it seems that the registered subscribers (maybe also publishers) are not correctly shutdown.  This leads to a topic still showing the rtt_environment as a subscriber to the topic after unloading the component that had a stream to ros.

I analyzed a little and it appears that the Channel for a publisher gets its destructor called because it is shared pointer https://github.com/orocos/rtt_ros_integration/blob/indigo-devel/rtt_roscomm/include/rtt_roscomm/rtt_rostopic_ros_msg_transporter.hpp#L247

I think one should also call shutdown on the ros_pub in the destructor after it is removed from the instance  https://github.com/orocos/rtt_ros_integration/blob/indigo-devel/rtt_roscomm/include/rtt_roscomm/rtt_rostopic_ros_msg_transporter.hpp#L129  

the same for the ros_sub in its destructor. I started to implement that, but the biggest problem I faced is that the destructor of the Channel for a subscriber was never called https://github.com/orocos/rtt_ros_integration/blob/indigo-devel/rtt_roscomm/include/rtt_roscomm/rtt_rostopic_ros_msg_transporter.hpp#L220
due to the msg transport creating a standard pointer and not a shared_ptr for it https://github.com/orocos/rtt_ros_integration/blob/indigo-devel/rtt_roscomm/include/rtt_roscomm/rtt_rostopic_ros_msg_transporter.hpp#L258  Not sure if the fact that the return type is a shared_ptr counts, because I added prints in the destructor and they are not printed.

I now know this is a header that has a larger implication and I struggled to understand that I had to rebuild the typekits for the ros message I wanted to test this cleanup for. However, even after making it a shared pointer it did still not enter the destructor.

What am I missing ?

guihomework:
completely fixes #61

### DIFF
--- a/rtt_roscomm/include/rtt_roscomm/rtt_rostopic_ros_msg_transporter.hpp
+++ b/rtt_roscomm/include/rtt_roscomm/rtt_rostopic_ros_msg_transporter.hpp
@@ -254,7 +254,7 @@ namespace rtt_roscomm {
       }
       else {
         if (!buf) return base::ChannelElementBase::shared_ptr();
-        tmp = new RosSubChannelElement<T>(port,policy);
+        tmp = base::ChannelElementBase::shared_ptr(new RosSubChannelElement<T>(port,policy));
         tmp->setOutput(buf);
         return tmp;
       }


meyerj:
The member variable `tmp` is already a shared pointer (actually a `boost::intrusive_ptr`) in [rtt_rostopic_ros_msg_transporter.hpp:257](https://github.com/orocos/rtt_ros_integration/pull/75/files#diff-b7dd3168b5e27cadacbc7d8c54c72e8dL257), so the change in this PR should actually not make a difference.

Only [intrusive_ptr](http://www.boost.org/doc/libs/1_61_0/libs/smart_ptr/intrusive_ptr.html)s allow assignment from raw pointers like in this case. [shared_ptr](http://www.boost.org/doc/libs/1_61_0/libs/smart_ptr/shared_ptr.htm)s with an external reference counter indeed would have to be constructed explicitly before the assignment to `tmp` (or using [make_shared](http://www.boost.org/doc/libs/1_61_0/libs/smart_ptr/make_shared.html)).

Did you observe a difference after having applied this patch? Actually the test added in https://github.com/orocos/rtt_ros_integration/pull/66 (only in branch toolchain-2.9) should verify that the subscriber was actually shutdown after the port has been disconnected. The toolchain-2.9 branch is fully backwards-compatible and you can also use it for toolchain 2.8 in ROS indigo.

guihomework:
Dear @meyerj  I will re do my own tests on 2.8.3 first and then on 2.9 if you say it should not change anything, because without my patch on 2.8.3 I could still see the rtt node being displayed as subscriber with rostopic info.
In order to test the code change I know I have to clean completely my build space since the typekits must be rebuilt to integrate the changes in this header file.
stay tuned

guihomework:
@meyerj  
After cleaning my build space AND devel space (using catkin tools) it this time worked. 
I had cleaned only my build space only , applying or removing my patch and I could definitely see changes, now with the devel space cleaned as well your patch is sufficient. I suppose the fact that code is in headers affects building with catkin somehow.

sorry for the noise.

meyerj:
> I suppose the fact that code is in headers affects building with catkin somehow.

Yeah, that's kind of tricky and the main reason why https://github.com/orocos-toolchain/rtt/pull/85 has been introduced in toolchain 2.9. Actually you can keep your build-space to save some time, but you should always clear the install-space (and devel-space in case you are using catkin) before rebuilding without this include directories patch because otherwise your compiler will prefer the installed headers from previous builds.

guihomework:
Okay, it makes sense now.
I am saving so much time building my workspace without rtt-ros-integration (since the released version works for me). 
thanks for the nice support.

PR #76, opened 2016-08-30
rtt_rosparam extra operations

ahoarau:
This PR adds some operations for ops scripts for the standard types : 
- bool --> getBool, setBool
- int --> getInt, setInt
- double --> getDouble, setDouble
- float 
  [rosparam_test.ops.zip](https://github.com/orocos/rtt_ros_integration/files/445319/rosparam_test.ops.zip)

--> getFloat, setFloat
- string --> getString, setString

and the according std::vector<T> : getVectorOfDouble ...

``` ruby
#!/usr/bin/env deployer
require("print")
import("rtt_rosnode")
import("rtt_rosparam")
loadService("this","rosparam")

# Experience parameters
var string robot_name
var double some_double
var bool use_sim_time
var array gains

# get them form rosparam
rosparam.getString("robot_name",robot_name)
rosparam.getDouble("~super_double",some_double)
rosparam.getBool("/use_sim_time",use_sim_time)
rosparam.getVectorOfDouble("super_gains",gains)

```

Test:

``` ruby
#!/usr/bin/env deployer
require("print")
import("rtt_rosnode")
import("rtt_rosparam")
loadService("this","rosparam")

var array s = array(1.0,2.0,3.0,4.4)
var array r

rosparam.setVectorOfDouble("/test_vec",s)
rosparam.getVectorOfDouble("/test_vec",r)

if(r[0] == 1.0 && r[1] == 2.0 && r[2] == 3.0 && r[3] == 4.4) then
    print.ln("Vector Success ")
else
    print.ln("Vector Failed ")

var string name = "orocos"
var string name_rec

rosparam.setString("/test_string",name)
rosparam.getString("/test_string",name_rec)

if(name_rec == "orocos") then
    print.ln("String Success --> "+name_rec)
else
    print.ln("String Failed ")
```

### DIFF
--- a/rtt_rosparam/include/rtt_rosparam/rosparam.h
+++ b/rtt_rosparam/include/rtt_rosparam/rosparam.h
@@ -36,7 +36,7 @@ namespace rtt_rosparam {
       setRelative("setRelative"),
       setAbsolute("setAbsolute"),
       setPrivate("setPrivate"),
-      setComponentPrivate("setComponentPrivate")
+      setComponentPrivate("setComponentPrivate"),
       setComponentRelative("setComponentRelative")
     {
       this->addOperationCaller(getAllRelative);
--- a/rtt_rosparam/include/rtt_rosparam/rosparam.h
+++ b/rtt_rosparam/include/rtt_rosparam/rosparam.h
@@ -4,6 +4,64 @@
 #include <rtt/RTT.hpp>
 #include <rtt/Property.hpp>
 
+// #include <Eigen/Dense>
+// #define eigen_matrix_xd Eigen::Matrix<double,Eigen::Dynamic,1>
+// #define eigen_matrix_xf Eigen::Matrix<float,Eigen::Dynamic,1>
+
+#ifndef ADD_ROSPARAM_SERVICE_CONSTRUCTOR
+#define ADD_ROSPARAM_SERVICE_CONSTRUCTOR(return_type_str) \
+  ,get##return_type_str("get"#return_type_str) \
+  ,set##return_type_str("set"#return_type_str) \
+  ,get##return_type_str##Relative ("get"#return_type_str"Relative") \
+  ,set##return_type_str##Relative ("set"#return_type_str"Relative") \
+  ,get##return_type_str##Absolute ("get"#return_type_str"Absolute") \
+  ,set##return_type_str##Absolute ("set"#return_type_str"Absolute") \
+  ,get##return_type_str##Private ("get"#return_type_str"Private") \
+  ,set##return_type_str##Private ("set"#return_type_str"Private") \
+  ,get##return_type_str##ComponentPrivate ("get"#return_type_str"ComponentPrivate") \
+  ,set##return_type_str##ComponentPrivate ("set"#return_type_str"ComponentPrivate") \
+  ,get##return_type_str##ComponentRelative ("get"#return_type_str"ComponentRelative") \
+  ,set##return_type_str##ComponentRelative ("set"#return_type_str"ComponentRelative") \
+  ,get##return_type_str##ComponentAbsolute ("get"#return_type_str"ComponentAbsolute") \
+  ,set##return_type_str##ComponentAbsolute ("set"#return_type_str"ComponentAbsolute")
+#endif
+
+#ifndef ADD_ROSPARAM_OPERATION_CALLER
+#define ADD_ROSPARAM_OPERATION_CALLER(return_type_str) \
+  this->addOperationCaller(get##return_type_str); \
+  this->addOperationCaller(set##return_type_str); \
+  this->addOperationCaller(get##return_type_str##Relative); \
+  this->addOperationCaller(set##return_type_str##Relative); \
+  this->addOperationCaller(get##return_type_str##Absolute); \
+  this->addOperationCaller(set##return_type_str##Absolute); \
+  this->addOperationCaller(get##return_type_str##Private); \
+  this->addOperationCaller(set##return_type_str##Private); \
+  this->addOperationCaller(get##return_type_str##ComponentPrivate); \
+  this->addOperationCaller(set##return_type_str##ComponentPrivate); \
+  this->addOperationCaller(get##return_type_str##ComponentRelative); \
+  this->addOperationCaller(set##return_type_str##ComponentRelative); \
+  this->addOperationCaller(get##return_type_str##ComponentAbsolute); \
+  this->addOperationCaller(set##return_type_str##ComponentAbsolute);
+#endif
+
+#ifndef DECLARE_ROSPARAM_OPERATION_CALLER
+#define DECLARE_ROSPARAM_OPERATION_CALLER(return_type_str, return_type) \
+  RTT::OperationCaller<bool(const std::string &, return_type &)>        get##return_type_str; \
+  RTT::OperationCaller<void(const std::string &, const return_type &)>  set##return_type_str; \
+  RTT::OperationCaller<bool(const std::string &, return_type &)>        get##return_type_str##Relative; \
+  RTT::OperationCaller<void(const std::string &, const return_type &)>  set##return_type_str##Relative; \
+  RTT::OperationCaller<bool(const std::string &, return_type &)>        get##return_type_str##Absolute; \
+  RTT::OperationCaller<void(const std::string &, const return_type &)>  set##return_type_str##Absolute; \
+  RTT::OperationCaller<bool(const std::string &, return_type &)>        get##return_type_str##Private; \
+  RTT::OperationCaller<void(const std::string &, const return_type &)>  set##return_type_str##Private; \
+  RTT::OperationCaller<bool(const std::string &, return_type &)>        get##return_type_str##ComponentPrivate; \
+  RTT::OperationCaller<void(const std::string &, const return_type &)>  set##return_type_str##ComponentPrivate; \
+  RTT::OperationCaller<bool(const std::string &, return_type &)>        get##return_type_str##ComponentRelative; \
+  RTT::OperationCaller<void(const std::string &, const return_type &)>  set##return_type_str##ComponentRelative; \
+  RTT::OperationCaller<bool(const std::string &, return_type &)>        get##return_type_str##ComponentAbsolute; \
+  RTT::OperationCaller<void(const std::string &, const return_type &)>  set##return_type_str##ComponentAbsolute;
+#endif
+
 namespace rtt_rosparam {
 
   class ROSParam : public RTT::ServiceRequester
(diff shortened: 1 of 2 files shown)


rtt_rosparam/include/rtt_rosparam/rosparam.h
```diff
@@ -53,31 +59,36 @@ namespace rtt_rosparam {
...
       this->addOperationCaller(setComponentPrivate);
+      this->addOperationCaller(setComponentRelative);
     }
 
     typedef enum  {
       RELATIVE, //! Relative resolution:  "name" -> "name"
       ABSOLUTE, //! Absolute resolution:  "name" -> "/name"
       PRIVATE,  //! Private resolution:   "name" -> "~name"
-      COMPONENT //! Component resolution: "name" -> "~COMPONENT_NAME/name"
+      COMPONENT_PRIVATE, //! Component resolution: "name" -> "~COMPONENT_NAME/name"
```
meyerj, line 156: Maybe the `COMPONENT` constant should stay, for backwards compatibility. We could solve this by adding

```
COMPONENT = COMPONENT_PRIVATE
```

at the end of the enumerator list.

rtt_rosparam/src/rtt_rosparam_service.cpp
```diff
@@ -12,6 +12,29 @@
 
 #include <ros/ros.h>
 
+#ifndef DECLARE_ROSPARAM_GETTER_SETTER
+#define DECLARE_ROSPARAM_GETTER_SETTER(function_name,return_type) \
```
meyerj: This macro could be replaced by template functions.

meyerj:
Some minor comments, otherwise looks good and useful to me. Thanks!

We should add a `rtt_rosparam_test` in the `tests/` subfolder of this repository which might include your Orocos snippet. This would probably require some minimal unit test framework for Orocos scripting if tests are supposed to do more than pure syntax checking, but that is off topic here and should be considered for upstream RTT.

rtt_rosparam/src/rtt_rosparam_service.cpp
ahoarau: but then how to declare all the addOperations ? does it support function overloading ?

meyerj: Something like this should work (untested):

``` cpp

#ifndef ADD_ROSPARAM_OPERATION
#define ADD_ROSPARAM_OPERATION(function_name, return_type) \
  this->addOperation("get"#function_name, &ROSParamService::getParamImpl< return_type >, this).doc("Get a " #return_type " from ROS parameter server"); \
  this->addOperation("set"#function_name, &ROSParamService::setParamImpl< return_type >, this).doc("Set a " #return_type " in the ROS parameter server");
#endif

class ROSParamService : public RTT::Service
{
public:
  ROSParamService(RTT::TaskContext *owner)
    : RTT::Service("rosparam", owner)
  {
    // [...]

    ADD_ROSPARAM_OPERATION(String, std::string)
    ADD_ROSPARAM_OPERATION(Double, double)
    ADD_ROSPARAM_OPERATION(Float, float)
    ADD_ROSPARAM_OPERATION(Int, int)
    ADD_ROSPARAM_OPERATION(Bool, bool)

    // Vector parameters
    ADD_ROSPARAM_OPERATION(VectorOfString, std::vector<string>)
    ADD_ROSPARAM_OPERATION(VectorOfDouble, std::vector<double>)
  }

  // [...]

private:
  template <typename T> bool getParamImpl(const std::string& ros_param_name, T& value)
  {
      if (!ros::param::get(ros_param_name, value)) {
          RTT::log(RTT::Debug) << "ROS Parameter \"" << ros_param_name << "\" not found on the parameter server!" << RTT::endlog();
          return false;
      }
      return true;
  }

  template <typename T> void setParamImpl(const std::string& ros_param_name, const T& value)
  {
      ros::param::set(ros_param_name, value);
  }
};
```

Note that I also updated the debug log message to include the return type instead of the function name and change the name of the macro to `ADD_ROSPARAM_OPERATION`. The private method `setParamImpl()` could be omitted if `&ros::param::set` is bound directly for the operation , but I like the explicit variant more.

meyerj: Should we also add all the `*Relative`, `*Absolute`, `*Private`, ... variants for each type-specific getter/setter operation? Currently they would default to policy `Relative`. For the case a component wants to get/set a parameter in a component-private namespace, the only possible invocation is

``` cpp
rosparam.getString("~name_of_the_component/parameter", s)
rosparam.getString("~" + getName() + "/parameter", s)
```

and `getName()` is not one of the default operations for components. In comparison:

``` cpp
rosparam.getStringComponentPrivate("parameter", s)
```

I actually would like the terms "Local" and "Global" more instead of "Relative" and "Absolute" (for another PR).

ahoarau:
all right I added the template, it works. 

+1 for adding a test, I'll try to implement one.

Instead of `rosparam.getStringComponentPrivate("parameter", s)`, could we do something like this: (?)

``` ruby
rosparam.getPrivate.Bool("parameter", s)
rosparam.getRelativeComponent.Bool("parameter", s)
```

meyerj:
> Instead of rosparam.getStringComponentPrivate("parameter", s), could we do something like this: (?)

For Orocos scripting the sub-service solution might look nice, but in C++ it would just bloat the code:

```
this->provides("rosparam")->provides("getPrivate")->getOperation("Bool")
```

The `ROSParam` ServiceRequester could partially solve that, but it is just provided for convenience and would either have the same problem or require an overridden [connectTo()](http://orocos-toolchain.github.io/devel/documentation/rtt/toolchain-2.9/api/html/classRTT_1_1ServiceRequester.html#a22bdc9b90ca3afb93b02251093d07c2d) method.

Unfortunately there is no operation overloading in RTT, so that the argument type could be deducted automatically and does not have to be known in advance, like for saving and loading property values. (It might be quite straight-forward to add this feature by allowing operation implementations to accept `RTT::base::DataSourceBase` arguments.)

For now I personally like the approach to add `2*m*n` operations for `n` types and `m` resolution policies most, following the

```
(get|set)(Relative|Absolute|Private|...)(Bool|Int|Float|Double|...)
```

pattern. An alternative would be to add a required third argument to the getter/setter operations, like for `get()` and `set()`.

Thanks for updating the PR and testing the proposed changes!

ahoarau:
Do you prefer the `2*m*n` lines or can I use `boost::preprocessor` ?

meyerj:
> Do you prefer the 2_m_n lines or can I use boost::preprocessor ?

No objections against Boost Preprocessor in this case. Or simply add all the resolution policy variants to the `ADD_ROSPARAM_OPERATION()` macro. Whatever is easier and improves readability.

ahoarau:
All right I added the following functions : 

```
(set | get)(Private | Relative | Absolute)
(set | get)Component(Private | Relative | Absolute)
```

We need to update the README also in the PR.

I added a few more tests : 

``` ruby
#!/usr/bin/env deployer
require("print")
import("rtt_rosnode")
import("rtt_rosparam")
loadService("this","rosparam")

var array s = array(1.0,2.0,3.0,4.4)
var array r

rosparam.setVectorOfDouble("/test_vec",s)
rosparam.getVectorOfDouble("/test_vec",r)

if(r[0] == 1.0 && r[1] == 2.0 && r[2] == 3.0 && r[3] == 4.4) then
    print.ln("Vector Success ")
else
    print.ln("Vector Failed ")

var string name = "orocos"
var string name_rec

rosparam.setString("~test_string",name)
rosparam.getStringPrivate("test_string",name_rec)

if(name_rec == "orocos") then
    print.ln("String private Success --> "+name_rec)
else
    print.ln("String private Failed ")

rosparam.setString("/test_string/absolute",name)
rosparam.getStringAbsolute("test_string/absolute",name_rec)

if(name_rec == "orocos") then
    print.ln("String absolute Success --> "+name_rec)
else
    print.ln("String absolute Failed ")

rosparam.setString("test_string",name)
rosparam.getStringRelative("test_string",name_rec)

if(name_rec == "orocos") then
    print.ln("String relative Success --> "+name_rec)
else
    print.ln("String relative Failed ")

var double a = 1.1
var double b = 0.0

rosparam.setDoubleComponentPrivate("test_double",a)
rosparam.getDoubleComponentPrivate("test_double",b)

if(a == b && b == 1.1) then
    print.ln("Double component private Success --> "+double(b))
else
    print.ln("Double component private Failed ")

```

ahoarau:
I also found a trick for Eigen Vectors, what do you think ?

Ex :

``` ruby
var array s = array(1.0,2.0,3.0,4.4)
rosparam.setVectorOfDouble("/test_vec",s)
var eigen_vector a
rosparam.getEigenVectorXd("/test_vec",a)
```

ahoarau:
1) I added the rosparam service with all the new operations 
2) I did not add eigen vectors operations because it needs to add <Eigen/Dense> in rosparam.h 
3) I think I did something wrong with the last commit

ahoarau:
All right the commit seems to be ok now. 

Do you think we should add Eigen in rosparam.h ?

meyerj:
I would prefer not to add explicit Eigen support to rtt_rosparam because it does not scale.

Ideally the problem of storing complex/non-standard types on the parameter server is solved by the respective typekit ([eigen_typekit](https://github.com/orocos/rtt_geometry/tree/indigo-devel/eigen_typekit) in this case) and RTT's type composition/decomposition mechanism, which already works fine for most typekits and also dynamic-sized Eigen vectors (at least I have not verified it for matrix types) if the source or target is a property. With the new operations, where the value is passed to or returned by the operation call directly, the function of rtt_rosparam is actually reduced to name resolution only. In this case people could directly call the roscpp API or a free helper functions instead. So maybe it is a better option to expose the `resolvedName()` method as an operation and leave the conversion to the user?

Or add operations that accept a `DataSourceBase` pointer and templated convenience wrappers in the service requester that create a temporary `ReferenceDataSource`. The latter should also work for all types that have a typekit with proper composition/decomposition support, like for properties.

meyerj:
> I would prefer not to add explicit Eigen support to rtt_rosparam because it does not scale.

Sorry, I completely misunderstood your question. rtt_rosparam already had special functions for Eigen types before. They were just not exposed in the rtt_rosparam.h service requester header and therefore the dependency on Eigen was not transitive. So please simply ignore what I have written before. (De)composition in rtt_rosparam is currently only supported for struct-like types, not for vectors.

I think in this case it is acceptable to also export Eigen as a dependency to other packages. I created a patch and at least added a preprocessor macro that allows to disable Eigen operations for the service requester with a one-line change, but the default configuration would include them.

I added some additional patches in branch [kuka-isir-toolchain-2.9](https://github.com/orocos/rtt_ros_integration/tree/kuka-isir-toolchain-2.9):
- https://github.com/kuka-isir/rtt_ros_integration/compare/toolchain-2.9...orocos:kuka-isir-toolchain-2.9 (partial diff)
- https://github.com/orocos/rtt_ros_integration/compare/toolchain-2.9...kuka-isir-toolchain-2.9 (full diff)

The last commit adds a new package rtt_rospack_tests with a quite complete unit test for rtt_rosparam. This test revealed some of the problems that have been fixed in the other commits.

I will wait a few days for review, especially because of the potentially API-breaking change in https://github.com/orocos/rtt_ros_integration/commit/e97ed8428be26bca75cdbec080b48e2b0efd73c1, but after that I think its good to merge. Thanks again for the contribution!

PR #83, opened 2017-01-27
Added a ROS transport plugin for primitive types

meyerj:
With only one additional light-weight transport plugin it is possible to connect/stream any Orocos port with a type that matches one of the built-in ROS types to a topic and map it to the corresponding [std_msgs](http://wiki.ros.org/std_msgs) type. I think this is really useful as it allows to rosify a lot of simple components without having to add a special port with a ROS message type.

Supported types and mappings:
- `bool` ([std_msgs/Bool](http://docs.ros.org/api/std_msgs/html/msg/Bool.html))
- `float` ([std_msgs/Float32](http://docs.ros.org/api/std_msgs/html/msg/Float32.html))
- `double` ([std_msgs/Float64](http://docs.ros.org/api/std_msgs/html/msg/Float64.html))
- `int8_t` ([std_msgs/Int8](http://docs.ros.org/api/std_msgs/html/msg/Int8.html))
- `int16_t` ([std_msgs/Int16](http://docs.ros.org/api/std_msgs/html/msg/Int16.html))
- `int32_t` ([std_msgs/Int32](http://docs.ros.org/api/std_msgs/html/msg/Int32.html))
- `int64_t` ([std_msgs/Int64](http://docs.ros.org/api/std_msgs/html/msg/Int64.html))
- `ros::Duration` ([std_msgs/Duration](http://docs.ros.org/api/std_msgs/html/msg/Duration.html))
- `ros::Time` ([std_msgs/Time](http://docs.ros.org/api/std_msgs/html/msg/Time.html))
- `uint8_t` ([std_msgs/Int8](http://docs.ros.org/api/std_msgs/html/msg/Int8.html))
- `uint16_t` ([std_msgs/UInt16](http://docs.ros.org/api/std_msgs/html/msg/UInt16.html))
- `uint32_t` ([std_msgs/UInt32](http://docs.ros.org/api/std_msgs/html/msg/UInt32.html))
- `uint64_t` ([std_msgs/UInt64](http://docs.ros.org/api/std_msgs/html/msg/UInt64.html))
- `std::string` ([std_msgs/String](http://docs.ros.org/api/std_msgs/html/msg/String.html))
- `RTT::rt_string` ([std_msgs/String](http://docs.ros.org/api/std_msgs/html/msg/String.html))

44c6d0b is a bit more complex and adds a type transporter for Orocos' `array` type, which is a `std::vector<double>` and maps to [std_msgs/Float64MultiArray](http://docs.ros.org/api/std_msgs/html/msg/Float64MultiArray.html) in ROS. The implementation requires a special adapter class.

f61f355 removes the `using namespace RTT` directive from rtt_rostopic_ros_msg_transporter.hpp and is actually unrelated. This header is typically only included in transport plugins and there were probably no side effects, but... who knows.

I do not see any compatibility problems with backporting these patches to `indigo-devel` and release it in indigo and jade, too.

We could do something similar for the services in [rtt_std_srvs](https://github.com/orocos/rtt_ros_integration/tree/toolchain-2.9/typekits/rtt_std_srvs) (Empty, Trigger, SetBool) and map them to one of the following operation signatures so that they can be called from other ROS nodes without adaptation of the component implementation:
```
std_srvs/Empty:
- void empty()

std_srvs/SetBool:
- bool setBool(bool, std::string &message_out)
- bool setBool(bool) // response.message will be empty
- std::string setBool(bool) // response.success = true
- void setBool(bool) // response.success = true and response.message will be empty

std_srvs/Trigger:
- bool trigger(std::string &message_out)
- bool trigger() // response.message will be empty
- std::string trigger() // response.success = true
```

### DIFF
--- a/rtt_ros/src/orocos/types/ros_primitives_typekit_plugin.cpp
+++ b/rtt_ros/src/orocos/types/ros_primitives_typekit_plugin.cpp
@@ -86,7 +86,7 @@ namespace ros_integration {
     void loadStringTypes();
 
     std::string ROSPrimitivesTypekitPlugin::getName(){
-	    return std::string("ros-")+"primitives";
+        return "ros-primitives";
     }
  
     bool ROSPrimitivesTypekitPlugin::loadTypes() {
--- a/rtt_roscomm/include/rtt_roscomm/rtt_rostopic_ros_msg_transporter.hpp
+++ b/rtt_roscomm/include/rtt_roscomm/rtt_rostopic_ros_msg_transporter.hpp
@@ -68,11 +68,27 @@
 namespace rtt_roscomm {
 
   /**
+   * An adapter that allows to overwrite the corresponding ROS type for a
+   * given Orocos type by specialization.
+   */
+  template <typename T>
+  struct RosMessageAdapter
+  {
+    typedef T OrocosType;
+    typedef T RosType;
+    static const RosType &toRos(const OrocosType &t) { return t; }
+    static const OrocosType &fromRos(const RosType &t) { return t; }
+  };
+
+  /**
    * A ChannelElement implementation to publish data over a ros topic
    */
   template<typename T>
   class RosPubChannelElement: public RTT::base::ChannelElement<T>, public RosPublisher
   {
+    typedef RosMessageAdapter<T> adapter;
+    typedef typename adapter::RosType RosType;
+
     char hostname[1024];
     std::string topicname;
     ros::NodeHandle ros_node;
@@ -122,9 +138,9 @@ namespace rtt_roscomm {
 
       // Handle private names
       if(topicname.length() > 1 && topicname.at(0) == '~') {
-        ros_pub = ros_node_private.advertise<T>(policy.name_id.substr(1), policy.size > 0 ? policy.size : 1, policy.init); // minimum 1
+        ros_pub = ros_node_private.advertise<RosType>(policy.name_id.substr(1), policy.size > 0 ? policy.size : 1, policy.init); // minimum 1
       } else {
-        ros_pub = ros_node.advertise<T>(policy.name_id, policy.size > 0 ? policy.size : 1, policy.init); // minimum 1
+        ros_pub = ros_node.advertise<RosType>(policy.name_id, policy.size > 0 ? policy.size : 1, policy.init); // minimum 1
       }
       act = RosPublishActivity::Instance();
       act->addPublisher( this );
@@ -190,7 +206,7 @@ namespace rtt_roscomm {
     bool write(typename RTT::base::ChannelElement<T>::param_t sample)
 #endif
     {
-      ros_pub.publish(sample);
+      ros_pub.publish(adapter::toRos(sample));
 #if RTT_VERSION_GTE(2,8,99)
       return RTT::WriteSuccess;
 #else
@@ -206,6 +222,9 @@ namespace rtt_roscomm {
   template<typename T>
   class RosSubChannelElement: public RTT::base::ChannelElement<T>
   {
+    typedef RosMessageAdapter<T> adapter;
+    typedef typename adapter::RosType RosType;
+
     std::string topicname;
     ros::NodeHandle ros_node;
     ros::NodeHandle ros_node_private;
@@ -253,10 +272,10 @@ namespace rtt_roscomm {
      * 
      * @param msg The received message
      */
-    void newData(const T& msg){
+    void newData(const RosType& msg){
       typename RTT::base::ChannelElement<T>::shared_ptr output = this->getOutput();
       if (output)
-          output->write(msg);
+          output->write(adapter::fromRos(msg));
     }
   };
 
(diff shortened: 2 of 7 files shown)


ahoarau:
This is great ! I haven't read the commits , but would it be possible to build the same for KDL messages ? Most of them are compatible with geometry_msgs

meyerj:
> This is great ! I haven't read the commits , but would it be possible to build the same for KDL messages ? > Most of them are compatible with geometry_msgs

There already is a package called [kdl_msgs](https://www.github.com/orocos/kdl_msgs) and typekit [rtt_kdl_msgs](https://www.github.com/orocos/rtt_kdl_msgs) which provide message and serialization traits for KDL types to be used as ROS messages and a ROS transport plugin for RTT, but with a focus on loss-less transport and as a drop-in replacement for KDL types as part of other messages.

We already discussed to add an option or provide an additional set of packages, like `(rtt_)kdl_geometry_msgs`, where the wire format is fully compatible to geometry_msgs. The main caveat is that for geometry_msgs all rotations would have to be converted to/from orientation quaternions, so publishing and subscribing a `KDL::Rotation` or derived types would have side effects.

ahoarau:
> We already discussed to add an option or provide an additional set of packages, like (rtt_)kdl_geometry_msgs, where the wire format is fully compatible to geometry_msgs. The main caveat is that for geometry_msgs all rotations would have to be converted to/from orientation quaternions, so publishing and subscribing a KDL::Rotation or derived types would have side effects.

Side effects in terms of performance ?

ahoarau:
```
We could do something similar for the services in rtt_std_srvs (Empty, Trigger, SetBool) and map them to one of the following operation signatures so that they can be called from other ROS nodes without adaptation of the component implementation:
```
I would very much need this indeed.

Services are somewhat useful, but having to write ROS wrappers or every operation that I want to convert into a service is a little cumbersome. 

Ideally the Service-to-operation connection would only be on the `ops` file, like the topics, to make a clean separation between ROS and OROCOS (the switch to ros2 would be less painful) . I wouldn't know how to start writing this automatic conversion. Do you have any plans for this ?

meyerj:
>> We already discussed to add an option or provide an additional set of packages, like (rtt_)kdl_geometry_msgs, where the wire format is fully compatible to geometry_msgs. The main caveat is that for geometry_msgs all rotations would have to be converted to/from orientation quaternions, so publishing and subscribing a KDL::Rotation or derived types would have side effects.

> Side effects in terms of performance ?

No, in terms of expected results. The transport would not be a pure transport, but has to convert from the rotation matrix representation inside `KDL::Rotation` to quaternions and/or vice-versa. Difficulties might arise from different frame conventions, small numeric errors during the conversion, flipped sign etc. That's more than what should be handled by a transport plugin.

> Ideally the Service-to-operation connection would only be on the ops file, like the topics, to make a clean separation between ROS and OROCOS (the switch to ros2 would be less painful) . I wouldn't know how to start writing this automatic conversion. Do you have any plans for this ?

Not at the moment. The approach would basically be as follows:
- Copy the [rtt_rosservice_proxies.cpp.in](https://github.com/orocos/rtt_ros_integration/blob/toolchain-2.9/rtt_roscomm/src/templates/service/rtt_rosservice_proxies.cpp.in) template from `rtt_roscomm` or the configured version from the build folder of `rtt_std_srvs` to `rtt_std_srvs/src`
- Do not call the macro in [CMakeLists.txt:6](https://github.com/orocos/rtt_ros_integration/blob/toolchain-2.9/typekits/rtt_std_srvs/CMakeLists.txt#L6), but add the necessary cmake code to compile the plugin from the copied and expanded version inside `rtt_std_srvs`
- Specialize or copy the `ROSServiceServerProxy<ROS_SERVICE_T>` class for the considered services and try to dynamic_cast the stored `proxy_operation_caller_` not only to one, but to multiple variants of [ProxyOperationCallerType](https://github.com/orocos/rtt_ros_integration/blob/toolchain-2.9/rtt_roscomm/include/rtt_roscomm/rtt_rosservice_proxy.h#L53) with different signatures. If one of them is ready, we can call the operation with the desired arguments and return values.
- Optionally, the [`ROSServiceServerProxyBase::connect(...)`](https://github.com/orocos/rtt_ros_integration/blob/toolchain-2.9/rtt_roscomm/include/rtt_roscomm/rtt_rosservice_proxy.h#L34) method could be made abstract with an implementation in the template class, so that the check which OperationCaller signature matches can already happen there. Depending on the result the `NodeHandle::advertiseService()` call can already register one out of n callback implementation that calls into the matching OperationCaller without the need for another dynamic_cast.

The documented usage at [Connecting RTT Operations to ROS Services](https://github.com/orocos/rtt_ros_integration/tree/toolchain-2.9/rtt_roscomm#connecting-rtt-operations-to-ros-services) still applies, only that `some_provided_service.some_operation` can be of any supported type for the requested ROS service and does not necessarily need to have an `bool(Request &, Response &)` signature.

PR #97, opened 2017-12-18
rtt_dynamic_reconfigure: Make PropertyBag passed to update callback to contain only changed properties.

achim-k:
Very often only a single parameter is changed via the `set_parameters` service. This change allows a custom implementation of the `updateProperties` operation to determine which properties have changed. The PropertyBag that is passed to the `updateProperties` operation contains only the properties that have changed.

Are there cases were this might not be desired? I can't think of any.

### DIFF
--- a/rtt_dynamic_reconfigure/include/orocos/rtt_dynamic_reconfigure/server.h
+++ b/rtt_dynamic_reconfigure/include/orocos/rtt_dynamic_reconfigure/server.h
@@ -635,16 +635,16 @@ private:
     {
         RTT::os::MutexLock lock(mutex_);
 
-        ConfigType new_config = config_;
-        traits::fromMessage(new_config, req.config, this);
-        traits::clamp(new_config, this);
-        uint32_t level = config_.__level__(new_config);
+        ConfigType msg_config;
+        traits::fromMessage(msg_config, req.config, this);
+        traits::clamp(msg_config, this);
+        uint32_t level = config_.__level__(msg_config);
 
         RTT::PropertyBag bag;
-        if (!updater()->propertiesFromConfig(new_config, level, bag)) return false;
+        if (!updater()->propertiesFromConfig(msg_config, level, bag)) return false;
         if (update_callback_.ready()) {
             if (!update_callback_(bag, level)) return false;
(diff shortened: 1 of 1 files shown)


rtt_dynamic_reconfigure/include/orocos/rtt_dynamic_reconfigure/server.h
```diff
@@ -635,32 +635,32 @@ class Server : public RTT::Service
...
+        updateConfigInternal(msg_config);
+        config_.__toMessage__(rsp.config);
         return true;
     }
 
     void updateConfigInternal(const ConfigType &config)
     {
         RTT::os::MutexLock lock(mutex_);
-        config_ = config;
+        RTT::updateProperties(config_, config);
```
meyerj, line 663: This is not allowed if `ConfigType` is something else than `AutoConfig` (which inherits from `RTT::PropertyBag`).

achim-k: Should there be something like `traits::updateConfigInternal(config_, msg_config, this);` then?

meyerj:
Your approach would work for `AutoConfig`, which can represent updates of only some parameters, but in traditional dynamic_reconfigure the user callback always get a full struct and the only way to know which of the parameters actually changed is via the `level` argument.

The update of the internal cached config could be implemented in the `traits` class, which is specialized for `AutoConfig`. Same for initializing `new_config` in `setConfigCallback()`: In this case you probably have to add an extra argument to `traits::fromMessage(...)` to pass the old config as a default value, which would be ignored by the specialized variant for `AutoConfig`.

meyerj:
This can also be tricky for composed types where a single property is composed of multiple dynamic reconfigure parameters. If you would only pass a partial PropertyBag to `propertiesFromConfig()` type composition might fail. An example would be a `KDL::Frame` property, that is decomposed into a group of 6 double parameters for x/y/z/roll/pitch/yaw, but with your patch you always would have to set the full frame information in a single callback or otherwise type composition will fail.

achim-k:
Ok. As this is more involved than I originally thought, I will close this PR and redo it some time later.

PR #106, opened 2018-07-06
add topicLatched to scripting

ahoarau:


meyerj:
Thanks. This patch could be backported to [indigo-devel](https://github.com/orocos/rtt_ros_integration/tree/indigo-devel), too.

Issue #104, opened 2018-06-05
Change the default number of spinner threads


ahoarau:
The default spinner thread count for `rtt_rosnode` [is set to 1 ](https://github.com/orocos/rtt_ros_integration/blob/toolchain-2.9/rtt_rosnode/src/ros_plugin.cpp#L82).

In roscpp code I see that setting it to zero gets us ths number of CPUs 
https://github.com/ros/ros_comm/blob/kinetic-devel/clients/roscpp/src/libros/spinner.cpp#L207:L209

Is there any reason why this is set to 1 and not 0 ?
I had an issue on xenomai not allowing the ROS spinner thread to send anything (xeno task was taking too much ressources), which was solved setting more threads.

meyerj:
I agree that the number of spinner threads has a significant effect on the overall behavior of a ROS application, including ROS-enabled Orocos processes. This is especially true if the process provides services, where a long-running callback actually blocks all subscriptions.

Increasing the number of default spinner threads would be one option to mitigate the problem, and unless someone would use ROS primitives with callbacks directly within the same process or `rtt_roscomm` service server proxies with `ClientThread` operations, there should be no noticeable side effect.

We had a similar problem in the past few weeks and I was thinking about the following alternative solutions, which eventually would scale better:
- a per subscriber or service server [CallbackQueue](http://wiki.ros.org/roscpp/Overview/Callbacks%20and%20Spinning) and associated Orocos activity
- a per component [CallbackQueue](http://wiki.ros.org/roscpp/Overview/Callbacks%20and%20Spinning) and associated Orocos activity, similar to the dispatcher threads of the CORBA or mqueue transport for the sending side. Similarly the [RosPublishActivity](https://github.com/orocos/rtt_ros_integration/blob/toolchain-2.9/rtt_roscomm/include/rtt_roscomm/rtt_rostopic_ros_publish_activity.hpp) could be instantiated per component instead of as a process-wide singleton.
- a customized implementation of [ros::CallbackQueue](http://docs.ros.org/api/roscpp/html/classros_1_1CallbackQueue.html) that instead of enqueuing a callback and waiting until a spinner thread polls it would call the callback immediately. This callback would directly push the message to the data or buffer object associated with the ROS stream connection, bypassing ROS' callback queue and the spinner threads completely. Orocos RTT already has all the primitives we need. On the other hand deserialization would already happen in roscpp's network thread and I am not sure whether this is a good idea.

ahoarau:
[...]

PR #116, opened 2019-02-13
Defaults the number of spinner threads

ahoarau:
To the number of CPUs available


meyerj:
Addresses #104.
LGTM, thanks!

Issue #101, opened 2018-02-28
ROS service specialization


ahoarau:
A follow up to the discussion in https://github.com/orocos/rtt_ros_integration/pull/83.

```
std_srvs/Empty:
- void empty()

std_srvs/SetBool:
- bool setBool(bool, std::string &message_out)
- bool setBool(bool) // response.message will be empty
- std::string setBool(bool) // response.success = true
- void setBool(bool) // response.success = true and response.message will be empty

std_srvs/Trigger:
- bool trigger(std::string &message_out)
- bool trigger() // response.message will be empty
- std::string trigger() // response.success = true
```

I started to implement this automatic conversion : 

```cpp

#include <rtt/RTT.hpp>
#include <rtt/plugin/ServicePlugin.hpp>
#include <rtt/internal/GlobalService.hpp>

#include <rtt_roscomm/rtt_rosservice_registry_service.h>
#include <rtt_roscomm/rtt_rosservice_proxy.h>

////////////////////////////////////////////////////////////////////////////////
#include <std_srvs/Empty.h>
#include <std_srvs/SetBool.h>
#include <std_srvs/Trigger.h>

////////////////////////////////////////////////////////////////////////////////
template<>
class ROSServiceServerProxy<std_srvs::Empty> : public ROSServiceServerProxyBase
{
public:
  //! All possible Operation callers for a ROS service server proxy
  typedef RTT::OperationCaller<bool(typename std_srvs::Empty::Request&, typename std_srvs::Empty::Response&)> ProxyOperationCallerType0;
  typedef RTT::OperationCaller<bool(void)> ProxyOperationCallerType1;
  typedef RTT::OperationCaller<void(void)> ProxyOperationCallerType2;

  /** \brief Construct a ROS service server and associate it with an Orocos
   * task's required interface and operation caller.
   */
  ROSServiceServerProxy(const std::string &service_name)
  : ROSServiceServerProxyBase(service_name)
  {
    // Construct operation caller
    proxy_operation_caller0_.reset(new ProxyOperationCallerType0("ROS_SERVICE_SERVER_PROXY0"));
    proxy_operation_caller1_.reset(new ProxyOperationCallerType1("ROS_SERVICE_SERVER_PROXY1"));
    proxy_operation_caller2_.reset(new ProxyOperationCallerType2("ROS_SERVICE_SERVER_PROXY2"));

    // Construct the ROS service server
    ros::NodeHandle nh;
    server_ = nh.advertiseService(
[...]

PR #123, opened 2019-05-07
Add wrappers for common operation signatures to rtt_std_srvs

meyerj:
Implements #101.

With this patch it is possible to create ROS service servers for some common operation signatures out of the box, without the need to write custom wrappers or to add extra operations with the ROS specific callback signature.

### Supported service types and associated signatures:
```
std_srvs/Empty:
- bool empty()                     // The service call fails if empty() returns false!
                                   // Use std_srvs/Trigger if the result should be returned as the response.
- void empty()

std_srvs/SetBool:
- bool setBool(bool, std::string &message_out)
- bool setBool(bool)               // response.message will be empty
- std::string setBool(bool)        // response.success = true
- void setBool(bool)               // response.success = true and response.message will be empty

std_srvs/Trigger:
- bool trigger(std::string &message_out)
- bool trigger()                   // response.message will be empty
- std::string trigger()            // response.success = true
```

The approach can be easily extended to other ROS service types.


tests/rtt_roscomm_tests/test/transport_tests.cpp
```diff
@@ -118,12 +118,23 @@ TEST(TransportTest, VectorTest)
 }
 
 static int callback_called = 0;
```
francisco-miguel-almeida, line 120: question: do we control in which thread the callbacks are made? If not, maybe this counter should be atomic. Not important right now, just curious.

francisco-miguel-almeida (approved):
LGTM, there was just a build failure due to pre-C++11 incompatibility. Issue is easily circumvented, however.

tests/rtt_roscomm_tests/test/transport_tests.cpp
meyerj: The callbacks are executed by a thread pool owned by a `ros::AsyncSpinner` in `rtt_rosnode`. They could execute in parallel, but the service calls in the unit tests are synchronous and block until the callback has been processed and the results are returned through the network stack and roscpp. So no concurrency.

francisco-miguel-almeida:
> LGTM, there was just a build failure due to pre-C++11 incompatibility. Issue is easily circumvented, however.

Was not quite as easily circumvented as it appeared at first, but now it works as intended.
We can address the SFINAE hackiness later.

ahoarau:
Very nice !

ahoarau:
very nice !

PR #124, opened 2019-05-08
Add wrappers for common operation signatures to rtt_std_srvs (follow-up)

meyerj:
Minor follow-up of #123:

@francisco-miguel-almeida Not sure why you moved the public methods of `ROSServiceServerOperationCaller<ROS_SERVICE_T>` to an extra section at the end of the class declaration in https://github.com/orocos/rtt_ros_integration/pull/123/commits/fd14e715f6eb4a80a16e981de0cbcef32dee7b38. At least the constructor and `proxy_operation_caller_` member should remain private.


francisco-miguel-almeida (approved):
Not sure why I left the public code split into two. I think I was concerned with undefined symbol errors on Windows from previous experience, but it's nothing to worry about here. Thanks for cleaning up.
LGTM.

PR #130, opened 2019-06-22
Xenomai 2/3 support update

ahoarau:
This goes with https://github.com/orocos-toolchain/rtt/pull/301 on rtt.
Builds fine on xenomai 2.x and 3.x.
I think I already asked, but isnt there a way to export the Findxxx.cmake from rtt ?


meyerj:
Thanks for your valuable contribution!

> I think I already asked, but isnt there a way to export the Findxxx.cmake from rtt ?

Yes. I do not see a problem with installing `FindXenomai.cmake` to `lib/cmake/orocos-rtt` (`${CONFIG_FILE_PATH}`) alongside other cmake configuration files in [rtt/CMakeLists.txt](https://github.com/orocos-toolchain/rtt/blob/master/CMakeLists.txt). Then any depending package can include the file after `OROCOS-RTT` has been found using

```cmake
find_package(OROCOS-RTT REQUIRED)
# or find_package(rtt_ros) or find_package(catkin REQUIRED COMPONENTS rtt_ros) in a ROS context

# Link against Xenomai POSIX skin to get
if(OROCOS_TARGET STREQUAL "xenomai")
  list(APPEND CMAKE_MODULE_PATH "${OROCOS-RTT_DIR}")
  find_package(Xenomai REQUIRED)

  # ...
endif()
```

I would not add `${OROCOS-RTT_DIR}` to the `CMAKE_MODULE_PATH` by default in the RTT cmake config or even include it implicitly from [orocos-rtt-config.cmake](https://github.com/orocos-toolchain/rtt/blob/master/orocos-rtt-config.cmake) to reduce the risk of conflicts with modules of the same name in other packages.

For me it would be acceptable to already remove the Xenomai cmake scripts from this repo and install the new file in https://github.com/orocos-toolchain/rtt/pull/301. The old scripts would probably not have worked anymore anyway, at least not for Xenomai 3.

_Side note:_ The same applies for other custom CMake config files, e.g. for CORBA. Some package redistribute them, e.g. [kdl_typekit](https://github.com/orocos/rtt_geometry/tree/toolchain-2.9/kdl_typekit/src/corba/config).

ahoarau:
Excellent, let's go with installing all those `cmake` files.
1. In order to make things coherent, the variables generated by `Find<PkgName>.cmake` should be `<PkgName>_LIBRARIES` etc. Here we have `XENOMAI_LIBRARIES` for backwards compatibility, but I'm fine with updating it.

2. Shall we change the name of the file ? I'm concerned people will load the ROS [cmake_modules](https://github.com/ros/cmake_modules) and executing the `FindXenomai.cmake` included.

ahoarau:
rebased from 2.10 + fix compilation on clang (c++ mangling)

PR #133, opened 2019-06-26
rtt_roscomm: fixed compilation of service plugins with Clang

meyerj:
Fixes a regression from https://github.com/orocos/rtt_ros_integration/pull/123, that caused build failures when compiling service plugins with Clang. Clang seems to handle SFINAE differently than gcc and compilation of generated service proxy plugins failed with errors like:

```cpp
In file included from rtt_diagnostic_msgs/diagnostic_msgs_service_proxies/rtt_rosservice_proxies.cpp:7:
rtt_roscomm/rtt_rosservice_proxy.h:99:30: error: implicit instantiation of undefined template 'ROSServiceServerOperationCallerWrapper<diagnostic_msgs::AddDiagnostics, 1>'
```


francisco-miguel-almeida (approved):
LGTM. +1 for using `boost::disable_if` for SFINAE, it makes the code easier to read now.

PR #136, opened 2019-10-03
Fix RTT_TF static stream initialization

Hugal31:
Correctly fill the fields of cp_static to create the stream.


meyerj (approved):
Thanks, good catch. Copy and paste...

PR #137, opened 2019-10-10
Fix rtt_actionlib simple action server

Hugal31:
Add a shutdown method to `RTTActionServer`, so it is possible to stop the timer, and moreover to make RTTSimpleActionServer::shutdown compile.

Also remove the `publish_feedback` flag to `RTTSimpleActionServer::start`, since `RTTActionServer::initialize` will be called anyway by `RTTActionServer::start` (see [the source code of actionlib::ActionServerBase::start](https://github.com/ros/actionlib/blob/e9c6c0dc58dd44a2d4e81c20e9dbd1103f44a62b/include/actionlib/server/action_server_base.h#L194-L200))


meyerj:
Looks good. Thanks for your contribution!

PR #138, opened 2019-10-25
Do not use ROS connection with rtt_rosclock::SimClockThread if time source is manual

Hugal31:
rtt_rosclock::SimClockThread used to instantiate a ros::NodeHandle event if the time source is `SIM_CLOCK_SOURCE_MANUAL`. This is a problem when one wants to write tests not relying on a ROS connection.

rtt_rosclock::SimClockThread now instantiate a ros::NodeHandle only if the time source is the /clock topic, and not the manual time source.


meyerj:
Sounds useful and the patch looks good. Thanks for your contribution!

PR #139, opened 2019-11-13
rtt_ros: Fix memory leak when parsing package xml file.

achim-k:
From http://www.xmlsoft.org/html/libxml-tree.html#xmlNodeGetContent:

> 
>```
>xmlChar *	xmlNodeGetContent	(const xmlNode * cur)
>```
> *Returns*: a new `#xmlChar *` or `NULL` if no content is available. It's up to the caller to free the memory with `xmlFree()`.


meyerj:
Looks good!

PR #147, opened 2020-08-12
rtt_rosparam: add direct access to Ros parameters

spd-intermodalics:
# Description

This PR adds support to access the value ROS parameter directly from one operation.

## Content

The next operations are added to the `rosparam` service:
* `getParam_type_(name)`: Gets the value of the ROS parameter named `name`. The namespace resolution is included in the `name` format. The value is returned instead of being assigned to an output parameter.
* `setParam_type_(name, value)`: Sets the value of a ROS parameter `name` to `value`. There is not need of an intermediary property. The name resolution is included in the `name` format.

### DIFF
--- a/rtt_rosparam/README.md
+++ b/rtt_rosparam/README.md
@@ -53,6 +53,7 @@ policy (see below).
 ##### Operations for getting single properties
 
 * **getParam(ros_name, rtt_name)** Get the ROS param **ros_name** and store it in the RTT property **rtt_name**. Use leaders like `~` and `/` for private and absolute resolution.
+* `getParam_type_(name)` Gets the value of a ROS param named `name`. The value returned is the default constructor if no `name` parameter can be found. Use leaders like `~` and `/` for private and absolute resolution.
 * **get(name ,policy)** Attempt to get the property named **name** (or populates the properties of a named RTT sub-service)
   from the ROS parameter namespace specified by **policy**.
 * **getRelative(name)**
@@ -63,6 +64,7 @@ policy (see below).
 ##### Operations for setting single properties
 
 * **setParam(ros_name, rtt_name)** Set the ROS param **ros_name** from the value in the RTT property **rtt_name**. Use leaders like `~` and `/` for private and absolute resolution.
+* `setParam_type(name, value)` Sets the value of the ROS param `name` to `value`. Use leaders like `~` and `/` for private and absolute resolution.
 * **set(name, policy)** Attempt to set the property named **name** (or stores the properties of a named RTT sub-service)
   in the ROS parameter namespace specified by **policy**.
 * **setRelative(name)**
@@ -135,7 +137,7 @@ class MyComponent : public RTT::TaskContext {
         all_params_found &= rosparam->getAbsolute("robot_description");
 
         // Get the ROS parameter "~publish_period"
-        all_params_found &= rosparam->getPrivate("publish_priod");
+        all_params_found &= rosparam->getPrivate("publish_period");
       }
       
       return all_params_found;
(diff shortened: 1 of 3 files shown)


rtt_rosparam/README.md
```diff
@@ -53,6 +53,7 @@ policy (see below).
 ##### Operations for getting single properties
 
 * **getParam(ros_name, rtt_name)** Get the ROS param **ros_name** and store it in the RTT property **rtt_name**. Use leaders like `~` and `/` for private and absolute resolution.
+* `getParam_type_(name)` Gets the value of a ROS param named `name`. The value returned is the default constructor if no `name` parameter can be found. Use leaders like `~` and `/` for private and absolute resolution.
```
meyerj, line 56:
```suggestion
* `getParam<Type>(name)` Gets the value of a ROS param named `name` of type `<type>`. The value returned is the default constructor if no `name` parameter can be found. Use leaders like `~` and `/` for private and absolute resolution.
```

But it does not fit in this category because it does not get a parameter value and stores it in a property:
> ##### Operations for getting single properties

rtt_rosparam/README.md
```diff
@@ -63,6 +64,7 @@ policy (see below).
 ##### Operations for setting single properties
 
 * **setParam(ros_name, rtt_name)** Set the ROS param **ros_name** from the value in the RTT property **rtt_name**. Use leaders like `~` and `/` for private and absolute resolution.
+* `setParam_type(name, value)` Sets the value of the ROS param `name` to `value`. Use leaders like `~` and `/` for private and absolute resolution.
```
meyerj, line 67:
```suggestion
* `setParam<Type>(name, value)` Sets the value of the ROS param `name` of type `<type>` to `value`. Use leaders like `~` and `/` for private and absolute resolution.
```

and it also needs to be moved to a separate section because it does not set a property.

meyerj (commented):
The naming conventions of the existing and new provided operations are inconsistent and confusing. There are already too many operations to basically do the same thing. And the documentation in [README.md](https://github.com/orocos/rtt_ros_integration/blob/4cd11e8a16964f30806cc97d9a3b623eba429229/rtt_rosparam/README.md) was not up-to-date, because it does not mention the new operations 

- `get<Type>(name, value)` and
- `set<Type>(name, value)`

that have already been added in https://github.com/orocos/rtt_ros_integration/pull/76. So basically some operations get the value of one or more ROS parameters and store it in respective RTT properties or vice-versa, and other get and set values directly without involving RTT properties.

Do we really need an additional `getParam<Type>()` method that returns the value directly, without the option to specify a default value? Such a method is error-prone because there is no way for the caller to know whether the returned value was retrieved from the parameter server or whether the call failed and the function returned a default-constructed value. The undocumented getter methods from #76 are almost equivalent and do not have that problem, because they take a non-const reference to store the value and return a boolean. And it is clear that we are dealing with parameters, because this is the `rosparam` service.

And `setParam<Type>()` is just an alias for `set<Type>()`. Why?

The prefix "~" alone is not sufficient to distinguish `PRIVATE` and `COMPONENT_PRIVATE` parameters. But that was already the case for `getParam(ros_name, rtt_name)` and `setParam(ros_name, rtt_name)` before. Not super important because the ROS parameter name can be passed freely, so the component name could be added by the caller.

The existing `getParam(ros_name, rtt_name)` operation actually copies the value of a ROS parameter to an existing RTT property and also returns a boolean. So maybe a name like `loadPropertyFromParam()` (and likewise `setParamFromProperty()`) would have been more descriptive when those operations have been added in https://github.com/orocos/rtt_ros_integration/commit/3d35760bb37dcbf5bf36d3860f790b5bb55b7dfc, but it should not be changed anymore for backwards compatibility. But the new operation `getParam<Type>(name)` proposed here would do something different and returns the value directly. That is kind of confusing. Maybe `getValue<Type>()` would be a better name, if it is decided to keep the new operations?

spd-intermodalics:
Because the feature already exists and in PR #76 and this only extends API without much advantage, I close it.

PR #148, opened 2020-08-12
rtt_rosparam: add a new DataSource linked to ROS parameter

spd-intermodalics:
# Description

This PR adds a new type or Orocos `DataSource` called `RTT::rosparam::RosParamDataSource<T>` derived from an `RTT::internal::AssignableDataSource<T>`.

The new data source interacts with the ROS parameter server to `get()` and `set()` source values. In this way, the data referred by the data source is linked to the ROS parameter.

## Contents

New class:
* `RTT::rosparam::RosParamDataSource<T>`: A data source that is linked to a ROS parameter

New operations in service `rosparam`:
* `addRosParamProperty_type__resolution_(name)`: It adds a `RTT::Property<T>` to the interface of the `TaskContext` owning the service `rosparam`. This property holds a data source of type `RTT::rosparam::RosParamDataSource<T>` linked to a ROS parameter. The property is of type `_type_` and is searched in the `_resolution_` namespace resolution policy.


rtt_rosparam/CMakeLists.txt
```diff
@@ -12,14 +12,24 @@ include_directories(
   ${catkin_INCLUDE_DIRS}
   ${EIGEN3_INCLUDE_DIRS})
 
-# build ROS param as a separate service
+# Add the data source header
+orocos_library(ros_param_data_source src/RosParamDataSource.cpp)
```
meyerj: Please follow the naming conventions of existing source files in this repository, e.g. `rtt_rosparam_data_source.cpp`.

rtt_rosparam/CMakeLists.txt
```diff
@@ -12,14 +12,24 @@ include_directories(
...
+  target_link_libraries(ros_param_data_source
```
meyerj:
```suggestion
target_link_libraries(ros_param_data_source
```

rtt_rosparam/CMakeLists.txt
```diff
@@ -12,14 +12,24 @@ include_directories(
...
 target_link_libraries(rtt_rosparam 
 #  ${OROCOS-RTT_RTT-MARSHALLING_LIBRARY} 
-  ${catkin_LIBRARIES})
+  ${catkin_LIBRARIES}
+  ros_param_data_source
+)
 
 # Generate install targets and pkg-config files
 orocos_install_headers(include/rtt_rosparam/rosparam.h)
+orocos_install_headers(include/rtt_rosparam/RosParamDataSource.hpp)
```
meyerj: Naming conventions: `rtt_rosparam_data_source.h`

rtt_rosparam/README.md
```diff
@@ -22,53 +22,63 @@ policy (see below).
...
+* `addRosParamProperty_type_Absolute(name)` Adds a property of type `_type_` with name `name` to the owning task context which is linked to the ROS parameter with the same name in a `Absolute` namespace resolution context.
+* `addRosParamProperty_type_Private(name)` Adds a property of type `_type_` with name `name` to the owning task context which is linked to the ROS parameter with the same name in a `Private` namespace resolution context.
+* `addRosParamProperty_type_ComponentPrivate(name)` Adds a property of type `_type_` with name `name` to the owning task context which is linked to the ROS parameter with the same name in a `ComponentPrivate` namespace resolution context.
+* `addRosParamProperty_type_ComponentAbsolute(name)` Adds a property of type `_type_` with name `name` to the owning task context which is linked to the ROS parameter with the same name in a `ComponentAbsolute` namespace resolution context.
+* `addRosParamProperty_type_ComponentRelative(name)` Adds a property of type `_type_` with name `name` to the owning task context which is linked to the ROS parameter with the same name in a `ComponentRelative` namespace resolution context.
```
meyerj: _Suggestion:_ Instead of adding new properties, it would be convenient to allow to reassign the data source of existing properties? Typically in a `TaskContext` properties are already added with
```
// method 1
this->addProperty("my_property", value_);  # reference member variable value_
```
or
```
// method 2
this->addProperty(my_property_);   # adds an existing property to the TaskContext interface
```

This code should be agnostic to ROS and not depend on `rtt_rosparam`. But for a certain deployment one or more properties of that component could be linked to ROS parameters, e.g. in a deployment script:
```
loadComponent("my_component", "MyComponent")
loadService("my_component", "rosparam")
my_component.rosparam.linkComponentPrivate("my_property")
// or
my_component.rosparam.linkAllComponentPrivate()
```

There is no way to change `T value_` such that it always refers to the latest value, but it would be possible to replace the internal data source of the existing property, to move the reference to the existing data source into the instance of `RosParamDataSource` and use that one to store the retrieved value instead of in `mcached_data_`. If the property is not linked, the internal data source could still be of type `ValueDataSource<T>()`.

Like that, at least after a call to
```cpp
this->getProperty("my_property")->evaluate();
```
the member variable `value_` is updated, and
```cpp
this->getProperty("my_property")->updated();
```
would store the latest value on the ROS parameter server.

I admit that the use cases are quite limited, because normal components that are not aware of `rtt_rosparam` also would not call those methods. For properties added with method 2, where the property itself is a member variable and not only its value, the normal set and get operations would implicitly synchronize with the parameter server.

rtt_rosparam/include/rtt_rosparam/RosParamDataSource.hpp
```diff
@@ -0,0 +1,221 @@
...
+
+    RosParamDataSource(std::string param_name_id, bool use_cached = true);
+
+    typename internal::DataSource<T>::result_t get() const
+    {
+      T return_value;
+      if (muse_cached_) {
+        if (!ros::param::getCached(mparam_name_, return_value)) {
+          // std::cerr << "The parameter " << mparam_name_ << " cannot be fetched." << std::endl;
+          throw(std::range_error("The parameter " + mparam_name_ + " cannot be fetched."));
```
meyerj:
```suggestion
          throw std::range_error("The parameter " + mparam_name_ + " cannot be fetched.");
```

`throw` is a C++ keyword, and not a function to call.

meyerj: Why [std::range_error](https://en.cppreference.com/w/cpp/error/range_error)?!? That does not fit in this case. Maybe you meant [std::out_of_range](http://www.cplusplus.com/reference/stdexcept/out_of_range/), which is what [std::map::at(key)](http://www.cplusplus.com/reference/map/map/at/) throws if `key` is not an element in the map. Or just [std::runtime_error](https://en.cppreference.com/w/cpp/error/runtime_error).

But actually a data source should not throw and always return a default value. Only `evaluate()` can fail. So users can call `evaluate()` if they want to know whether the parameter is set, and then get the value with `value()` or `rvalue()`. `get()` should just return a default-constructed value or the previous cached value if `ros::param::get()` fails.

Typically `get()` is just implemented as
```cpp
    typename internal::DataSource<T>::result_t get() const
    {
      (void) evaluate();
      return value();
    }
```
and the actual work of updating `mcached_data_` happens in `evaluate()`.

rtt_rosparam/include/rtt_rosparam/RosParamDataSource.hpp
```diff
@@ -0,0 +1,221 @@
...
+          return T();
+        }
+      }
+      return mcached_data_ = return_value;
+    }
+
+    typename internal::DataSource<T>::result_t value() const
+    {
+      return get();
```
meyerj: `value()` and `rvalue()` should not re-evaluate the data source, i.e. not trigger `ros::param::get()` or `ros::param::getCached()` again. They must return the same value if called repeatedly without `evaluate()` or `get()`, which implies `evaluate()`.

rtt_rosparam/include/rtt_rosparam/RosParamDataSource.hpp
```diff
@@ -0,0 +1,221 @@
...
+
+    // There is not referred element, no allocation exists for this data source and it is not an alias.
+    // But it needs to return something, because otherwise the .write() calls that require a rvalue()
+    // would fail. So we use an internal cache
+    typename internal::DataSource<T>::const_reference_t rvalue() const
+    {
+      return mcached_data_;
+    }
+
+    void set( typename internal::AssignableDataSource<T>::param_t t );
```
meyerj: What is the reason to define some (even non-trivial) methods directly in the class declaration, but others below? Methods defined in the declaration are inlined implicitly, but that is the only difference.

rtt_rosparam/include/rtt_rosparam/RosParamDataSource.hpp
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    typename internal::DataSource<T>::result_t value() const
+    {
+      return get();
+    }
+
+    // There is not referred element, no allocation exists for this data source and it is not an alias.
+    // But it needs to return something, because otherwise the .write() calls that require a rvalue()
+    // would fail. So we use an internal cache
```
meyerj: See comment above. `value()` and `rvalue()` are expected to not reevaluate the data source and just return the cached value.

rtt_rosparam/include/rtt_rosparam/RosParamDataSource.hpp
```diff
@@ -0,0 +1,221 @@
...
+    ~RosParamDataSource();
+
+    typedef boost::intrusive_ptr< RosParamDataSource<T> > shared_ptr;
+
+    RosParamDataSource(std::string param_name_id, bool use_cached = true);
+
+    typename internal::DataSource<T>::result_t get() const
+    {
+      T return_value;
+      if (muse_cached_) {
```
meyerj: Why is this even an option exposed to the user? What is the disadvantage of not using `ros::param::getCached()` unconditionally? I think it perfectly fits this use case unless there is a downside of parameter caching in roscpp that I am not aware of.

rtt_rosparam/include/rtt_rosparam/RosParamDataSource.hpp
```diff
@@ -0,0 +1,221 @@
...
+  //       return 0.0;
+  //     }
+  //   } else {
+  //     if (!ros::param::get(mparam_name_, return_value)) {
+  //       throw(std::range_error("The parameter " + mparam_name_ + " cannot be fetched."));
+  //       return 0.0;
+  //     }
+  //   }
+  //   return return_value;
+  // }
```
meyerj: remove dead code

rtt_rosparam/include/rtt_rosparam/RosParamDataSource.hpp
```diff
@@ -0,0 +1,221 @@
...
+
+  template<typename T>
+  RosParamDataSource<T>::~RosParamDataSource() {}
+
+  template<typename T>
+  RosParamDataSource<T>::RosParamDataSource( std::string param_name, bool use_cached)
+      : mparam_name_(param_name),
+        muse_cached_(use_cached),
+        mcached_data_(T())
```
meyerj:
```suggestion
        mcached_data_()
```

rtt_rosparam/include/rtt_rosparam/RosParamDataSource.hpp
```diff
@@ -0,0 +1,221 @@
...
+  {
+  }
+  template<typename T>
+  void RosParamDataSource<T>::set( typename internal::AssignableDataSource<T>::param_t t )
+  {
+    ros::param::set(mparam_name_, t);
```
meyerj: Must also update `mcached_data_` or `value()` and `rvalue()` would return an outdated value afterwards.
Could be implemented as
```cpp
template<typename T>
  void RosParamDataSource<T>::set( typename internal::AssignableDataSource<T>::param_t t )
  {
    set() = t;
    ros::param::set(mparam_name_, mcached_data_);
  }

File: rtt_rosparam/include/rtt_rosparam/RosParamDataSource.hpp [hunk 13]
```diff
@@ -0,0 +1,221 @@
...
+
+  // template <typename T>
+  // typename internal::AssignableDataSource<T>::reference_t RosParamDataSource<T>::set() {
+  //   // return mdummy_;
+  // }
+
+  template<typename T>
+  RosParamDataSource<T>* RosParamDataSource<T>::clone() const
+  {
+      return new RosParamDataSource<T>(this->mparam_name_);
```
meyerj: Unless `muse_cached` is dropped anyway, it needs to be preserved in clones.

rtt_rosparam/include/rtt_rosparam/RosParamDataSource.hpp
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  template<typename T>
+  RosParamDataSource<T>* RosParamDataSource<T>::copy( std::map<const base::DataSourceBase*, base::DataSourceBase*>& alreadyCloned ) const {
+      return const_cast<RosParamDataSource<T>*>(this); // no copy needed, data is outside.
```
meyerj: That is not true because the data source eventually caches the value of the property. It is safer to also clone the data source conditionally, like in [ValueDataSource<T>::copy()](https://github.com/orocos-toolchain/rtt/blob/600102e8be9c81905b20930e32d43b28244ab173/rtt/internal/DataSources.inl#L88). It has a subtle but important effect if the data source is evaluated in scripting functions, and reusing the same `mcached_data_` variable in all copies would not be thread-safe.

rtt_rosparam/include/rtt_rosparam/RosParamDataSource.hpp
```diff
@@ -0,0 +1,221 @@
...
+  } // namespace rosparam
+} //namespace RTT
+
+// /*
+//  * Extern template declarations for core data source types
+//  * (instantiated in DataSources.cpp)
+// */
+RTT_EXT_IMPL template class RTT::rosparam::RosParamDataSource< bool >;
+RTT_EXT_IMPL template class RTT::rosparam::RosParamDataSource< std::string >;
+
```
meyerj:
```suggestion
```

This snippet obviously has been copied from [rtt/internal/DataSource.hpp](https://github.com/orocos-toolchain/rtt/blob/600102e8be9c81905b20930e32d43b28244ab173/rtt/internal/DataSource.hpp#L240). It is not needed here. In case of `RTT::internal::DataSource<T>` those data sources are already used internally in within RTT, which is why they are declared `extern` for every user of RTT. But that is not the case here.

However, you could use extern template declarations for all supported ROS parameter types here and explicitly instantiate them in the source, but then the comment is wrong and the list of types is incomplete. Easier to define the template class as-is and let the compiler return an error if it is used with a template type argument which is not supported by `ros::param::get()` or `ros::param::set()`.

rtt_rosparam/src/RosParamDataSource.cpp
```diff
@@ -0,0 +1,88 @@
+/***************************************************************************
+  tag: Sergio Portoles  Thu Aug 06 14:30:00 CEST 2020  RosParamDataSource.cpp
+
+                        RosParamDataSource.cpp -  description
+                           -------------------
+    begin                : Thu Aug 06 2020
+    copyright            : (C) 2020 Peter Soetens
+    email                : peter.soetens@fmtc.be
```
meyerj: outdated copyright information

rtt_rosparam/include/rtt_rosparam/RosParamDataSource.hpp
```diff
@@ -0,0 +1,221 @@
...
+#include "ros/ros.h"
+#include "ros/param.h"
+#include "ros/names.h"
+
+#include <boost/function.hpp>
+#include <exception>
+
+namespace RTT
+{
```
meyerj: Better not add rosparam as a sub-namespace of `RTT`, and instead follow the default convention in ROS to use the package name `rtt_rosparam` as the C++ namespace, like in [rtt_rosparam/rosparam.h](https://github.com/orocos/rtt_ros_integration/blob/2568773e68378bb41e668e2cffc1bf251f3f3484/rtt_rosparam/include/rtt_rosparam/rosparam.h#L68) and other packages in this repository. So the fully-qualified class name would be `rtt_rosparam::RosParamDataSource<T>`. You would have to also fully qualify RTT types in the header, and can add a `using namespace RTT` directive in the source file.

rtt_rosparam/src/RosParamDataSource.cpp
```diff
@@ -0,0 +1,88 @@
...
+ * Explicit template instantiation for core data source types
+ */
+template class RTT::rosparam::RosParamDataSource< bool >;
+template class RTT::rosparam::RosParamDataSource< std::string >;
+
+
+// A possible application: new addRosParamProperty()
+//
+// /**
+//  * Adds a ROS parameter of any type as a property to this bag.
```
meyerj: Only types supported by the ROS parameter server, which is only `bool`, `int`, `std::string`, `double` and `float`, and respective vector types. It would be possible to adapt other types implicitly or use composition/decomposition, but that is not implemented in the current version of `RosParamDataSource<T>`.

rtt_rosparam/src/rtt_rosparam_service.cpp
```diff
@@ -239,6 +247,43 @@ class ROSParamService: public RTT::Service
   bool setParamComponentPrivate(const std::string &name) { return set(name, COMPONENT_PRIVATE); }
   bool setParamComponentRelative(const std::string &name) { return set(name, COMPONENT_RELATIVE); }
   bool setParamComponentAbsolute(const std::string &name) { return set(name, COMPONENT_ABSOLUTE); }
+
+  /**
+   * Adds a ROS parameter of any type as a property to this bag.
```
meyerj: Not of any type. See comment above.

rtt_rosparam/src/rtt_rosparam_service.cpp
```diff
@@ -239,6 +247,43 @@ class ROSParamService: public RTT::Service
...
+   * A Property is created which causes contents of the
+   * property always to be in sync
+   * with the contents of the ROS parameter.
+   * @param name The name of this property
+   * @return the Property object by reference, which you can further query or document.
+   */
+  template<typename T, ResolutionPolicy P>
+  Property<T>& addRosParamProperty( const std::string& name) {
+  const std::string resolved_name = resolvedName(name,P);
+  typename internal::AssignableDataSource<T>::shared_ptr datasource( new RTT::rosparam::RosParamDataSource<T>(resolved_name) );
```
meyerj: style: broken indentation

rtt_rosparam/src/rtt_rosparam_service.cpp
```diff
@@ -239,6 +247,43 @@ class ROSParamService: public RTT::Service
...
+    // if (this->getOwner() == nullptr) {
+    //   return nullptr;
+    // }
+    // if (this->getOwner()->properties() == nullptr) {
+    //   return nullptr;
+    // }
```
meyerj:
```suggestion
```

remove dead code
`rosparam` can only be loaded into a component, so `getOwner()` and `getOwner()->properties()` must not return `nullptr`. If they would, also other calls in this class would cause segfaults.

And this repo must compile without C++11 enabled, so please do not use `nullptr` or other C++1x features.

rtt_rosparam/src/rtt_rosparam_service.cpp
```diff
@@ -239,6 +247,43 @@ class ROSParamService: public RTT::Service
...
+   * An interface ready version of the addRosParamProperty() member function.
+   * The main difference is that it returns the Property instead of a reference
+   * to the Property.
+   * @param name The name of this property
+   * @return the Value o the Property.
+   */
+  template<typename T, ResolutionPolicy P>
+  RTT::Property<T> addRosParamPropertyIface( const std::string& name) {
+    return addRosParamProperty<T, P>(name);;
+  }
```
meyerj: Why is this needed? It does not make a difference for the caller. Why would you want to enforce that a copy is returned?

meyerj (changes requested):
First review pass and there are some change requests and some suggestions.

It should be mentioned (here and in the documentation) that properties with a `RosParamDataSource<T>` are not real-time safe, because evaluation and updates involve a call to roscpp and network communication with the ROS master. So they should only be used in non-real-time contexts, e.g. in the DeploymentComponent or an extra configuration component.

rtt_rosparam/README.md
meyerj: We agreed that the suggested behavior that replaces existing data sources is error-prone and should not be implemented, because the resulting behavior breaks method 1 above.

rtt_rosparam/README.md
```diff
@@ -72,6 +72,14 @@ policy (see below).
 
 ##### Operations to create properties linked to a ROS param
 
+The following operations of the `rosparam` service can be used to create Properties
+in the owner component that link to ROS parameters. When the properties are
+`evaluate()`-d through `get()` or `set()`, the ROS parameter server is queried.
+
+**IMPORTANT** Setting or getting the value of a ROS parameter through a
+Property is a non real-time safe operation and therefore it shouldn't be used
+in the `updateHook()` of a real-time periodic component.
```
meyerj:
```suggestion
from within a real-time component.
```

_Rationale:_
- Not every real-time component must be periodic.
- There are also other primitives that run in the component's thread, not only `updateHook()`. Examples are `OwnThread` operations, event port callbacks, scripted programs and state machines.

rtt_rosparam/README.md
```diff
@@ -22,53 +22,71 @@ policy (see below).
...
-* **getAllPrivate()** Attempt to get all properties of this component (and its sub-services)
+* `getAllPrivate()` Attempt to get all properties of this component (and its sub-services)
   from the ROS parameter server in the node's private namespace.
 
 ##### Operations for setting all properties
-* **setAll()** or **setAllComponentPrivate()** Stores all properties of this component (and its sub-services)
+* `setAll()` or `setAllComponentPrivate()` Stores all properties of this component (and its sub-services)
   on the ROS parameter server from the similarly-named property in the **COMPONENT**'s private namespace.
-* **setAllRelative()** Stores all properties of this component (and its sub-services)
+* `setAllRelative()`` Stores all properties of this component (and its sub-services)
```
meyerj:
```suggestion
* `setAllRelative()` Stores all properties of this component (and its sub-services)
```

rtt_rosparam/README.md
```diff
@@ -22,53 +22,71 @@ policy (see below).
...
+* `set(name, policy)` Attempt to set the property named `name` (or stores the properties of a named RTT sub-service)
+  in the ROS parameter namespace specified by `policy`.
+* `setRelative(name)`
+* `setAbsolute(name)`
+* `setPrivate(name)`
+* `setComponentPrivate(name)`
+
+##### Operations to create properties linked to a ROS param
+
+The following operations of the `rosparam` service can be used to create Properties
```
meyerj:
```suggestion
The following operations of the `rosparam` service can be used to create properties
```

rtt_rosparam/README.md
```diff
@@ -22,53 +22,71 @@ policy (see below).
...
+in the owner component that link to ROS parameters. When the properties are
+`evaluate()`-d through `get()` or `set()`, the ROS parameter server is queried.
+
+**IMPORTANT** Setting or getting the value of a ROS parameter through a
+Property is a non real-time safe operation and therefore it shouldn't be used
```
meyerj:
```suggestion
property is a non real-time safe operation and therefore it shouldn't be used
```

rtt_rosparam/include/rtt_rosparam/ros_param_data_source.hpp
```diff
@@ -0,0 +1,147 @@
...
+  /**
+   * Stores the full-qualified property name of the ROS parameter
+   */
+  std::string mparam_name_;
+  mutable typename internal::DataSource<T>::value_t mcached_data_;
+
+public:
+  /**
+   * Use shared_ptr.
+   */
```
meyerj: nit: Has this been copied from somewhere? It is not a very meaningful documentation...
```suggestion
```

rtt_rosparam/include/rtt_rosparam/ros_param_data_source.hpp
```diff
@@ -0,0 +1,147 @@
...
+    (void) evaluate();
+    return value();
+  }
+
+  bool evaluate() const
+  {
+    T return_value;
+    if (!ros::param::getCached(mparam_name_, return_value)) {
+      throw std::out_of_range("The parameter " + mparam_name_ + " cannot be fetched.");
+      return false;
```
meyerj: `throw` or `return false`, but not both. Line 96 is not reachable.

Exceptions are not real-time safe, so Orocos uses them sparely in general. This method is not real-time safe anyway. Still, `DataSourceBase::evaluate()` typically does not throw, and just returns false. You could additionally log something with `Error` severity, if you prefer.

```suggestion
      RTT::log(RTT::Error) << "The value of parameter " << mparam_name_ << " could not be fetched.".
      return false;
```

rtt_rosparam/include/rtt_rosparam/ros_param_data_source.hpp
```diff
@@ -0,0 +1,147 @@
...
+  {
+    set() = t;
+    ros::param::set(mparam_name_, t);
+  }
+
+  // There is not referred element, no allocation exists for this data source and it is not an alias.
+  typename internal::AssignableDataSource<T>::reference_t set()
+  {
+    return mcached_data_;
+  }
```
meyerj: The class is missing an override of [AssignableDataSource<T>::updated()](https://github.com/orocos-toolchain/rtt/blob/600102e8be9c81905b20930e32d43b28244ab173/rtt/base/DataSourceBase.hpp#L148-L153). It is important to call that method after changing the cached value via `set()` without arguments. For example:
```cpp
RosParamDataSource<std::string> ds("~param");
functionThatTakesAReference(ds->set());
ds->updated();
```

```suggestion
  void set( typename internal::AssignableDataSource<T>::param_t t )
  {
    mcached_data_ = t;
    updated();
  }

  typename internal::AssignableDataSource<T>::reference_t set()
  {
    return mcached_data_;
  }

  void updated() {
    ros::param::set(mparam_name_, mcached_data_);
  } 
```

rtt_rosparam/src/ros_param_data_source.cpp
```diff
@@ -0,0 +1,80 @@
...
+  // {
+  // }
+
+} // namespace rtt_rosparam
+
+/*
+ * Explicit template instantiation for core data source types
+ */
+template class rtt_rosparam::RosParamDataSource< bool >;
+template class rtt_rosparam::RosParamDataSource< std::string >;
```
meyerj: Please remove this source file. It is of no use without an [extern template class](https://isocpp.org/wiki/faq/cpp11-language-templates#extern-templates) declaration in the respective header, and a good use case where the header is included by users. Normally for RTT plugins that is not the case. The `RosParamDataSource<T>` template class is so light-weight that it's not worth the additional complexity.

rtt_rosparam/src/rtt_rosparam_service.cpp
```diff
@@ -239,7 +247,27 @@ class ROSParamService: public RTT::Service
...
+   * with the contents of the ROS parameter.
+   * @param name The name of this property
+   * @return the Property object by reference, which you can further query or document.
+   */
+  template<typename T, ResolutionPolicy P>
+  Property<T>& addRosParamProperty( const std::string& name) {
+    const std::string resolved_name = resolvedName(name,P);
+    typename internal::AssignableDataSource<T>::shared_ptr datasource( new rtt_rosparam::RosParamDataSource<T>(resolved_name) );
+    Property<T>* p = new Property<T>(name,"", datasource);
+    p->doc("Property generated by rosparam attached to parameter: " + resolved_name);
```
meyerj: _Optional:_
```suggestion
    Property<T>* p = new Property<T>(name, "ROS parameter " + resolved_name, datasource);
```

Not sure why you first construct the `Property<T>` with an empty description and then set it afterwards with `p->doc()`?

rtt_rosparam/src/rtt_rosparam_service.cpp
```diff
@@ -239,7 +247,27 @@ class ROSParamService: public RTT::Service
...
   bool setParamComponentRelative(const std::string &name) { return set(name, COMPONENT_RELATIVE); }
   bool setParamComponentAbsolute(const std::string &name) { return set(name, COMPONENT_ABSOLUTE); }
-};
+
+  /**
+   * Adds a ROS parameter of supported types by ROS param as a property to this
+   * bag.
+   * A Property is created which causes contents of the
+   * property always to be in sync
+   * with the contents of the ROS parameter.
```
meyerj:
```suggestion
   * The new property mirrors the value of the respective ROS parameter of the same name.
   * Setting the value of the property sets the ROS parameter, and evaluating it
   * loads the current value from the ROS parameter server.
```

rtt_rosparam/include/rtt_rosparam/ros_param_data_source.hpp
```diff
@@ -0,0 +1,147 @@
...
+
+  typename internal::DataSource<T>::result_t get() const {
+    (void) evaluate();
+    return value();
+  }
+
+  bool evaluate() const
+  {
+    T return_value;
+    if (!ros::param::getCached(mparam_name_, return_value)) {
```
meyerj: Avoid an unnecessary copy in line 98 below:
```suggestion
    if (!ros::param::getCached(mparam_name_, mcached_data_)) {
```

rtt_rosparam/src/rtt_rosparam_service.cpp
```diff
@@ -239,7 +247,27 @@ class ROSParamService: public RTT::Service
...
+   * @param name The name of this property
```
meyerj: ... and also the name of the linked ROS parameter, right?

```suggestion
   * @param name The name of the ROS parameter and property to be added
```

rtt_rosparam/CMakeLists.txt
```diff
@@ -12,14 +12,24 @@ include_directories(
...
+)
+
+  # build ROS param as a separate service
 orocos_service(rtt_rosparam src/rtt_rosparam_service.cpp)
 target_link_libraries(rtt_rosparam 
 #  ${OROCOS-RTT_RTT-MARSHALLING_LIBRARY} 
-  ${catkin_LIBRARIES})
+  ${catkin_LIBRARIES}
+  ros_param_data_source
+)
```
meyerj: With the unnecessary source file removed, this change must be reverted. As a template class, `RosParamDataSource<T>` is header-only.

meyerj (changes requested):

rtt_rosparam/include/rtt_rosparam/ros_param_data_source.hpp
```diff
@@ -115,15 +112,22 @@ class RosParamDataSource
   void set( typename internal::AssignableDataSource<T>::param_t t )
   {
     set() = t;
-    ros::param::set(mparam_name_, t);
+    updated();
   }
 
-  // There is not referred element, no allocation exists for this data source and it is not an alias.
+  // There is not referred element, no allocation exists for this data source
+  // and it is not an alias.
```
meyerj:
```suggestion
```

Not sure what you want to say with this sentence, but it is grammatically wrong.

rtt_rosparam/include/rtt_rosparam/ros_param_data_source.hpp
```diff
@@ -115,15 +112,22 @@ class RosParamDataSource
...
+  // After working with the reference, if something was assigned, the data
+  // source may need a call to updated() after set()
   typename internal::AssignableDataSource<T>::reference_t set()
   {
     return mcached_data_;
   }
 
+  void updated() {
+    ros::param::set(mparam_name_, mcached_data_);
```
meyerj: nit: style consistency
```suggestion
  void updated()
  {
    ros::param::set(mparam_name_, mcached_data_);
```

rtt_rosparam/src/rtt_rosparam_service.cpp
```diff
@@ -251,18 +251,18 @@ class ROSParamService: public RTT::Service
   /**
    * Adds a ROS parameter of supported types by ROS param as a property to this
    * bag.
-   * A Property is created which causes contents of the
-   * property always to be in sync
-   * with the contents of the ROS parameter.
-   * @param name The name of this property
+   * The new property mirror the value of the respective ROS parameter of the
```
meyerj:
```suggestion
   * The new property mirrors the value of the respective ROS parameter of the same
```

rtt_rosparam/include/rtt_rosparam/ros_param_data_source.hpp
```diff
@@ -0,0 +1,151 @@
+/***************************************************************************
+ *        This file is part of the OROCOS toolchain ROS project
+ *
+ *                  (C) 2020, Intermodalics BVBA
+ *
```
meyerj: Please align the license header style and wording with the latest updates in #154.

rtt_rosparam/include/rtt_rosparam/ros_param_data_source.hpp
```diff
@@ -0,0 +1,151 @@
...
+#include "rtt/internal/DataSourceTypeInfo.hpp"
+#include "rtt/internal/Reference.hpp"
+#include "rtt/Logger.hpp"
+
+#include "ros/ros.h"
+#include "ros/param.h"
+#include "ros/names.h"
+
+#include <boost/function.hpp>
+#include <exception>
```
meyerj: Please check which includes are actually required and clean it up. At least the last two are not needed, and [`ros/ros.h`](https://github.com/ros/ros_comm/blob/HEAD/clients/roscpp/include/ros/ros.h) is a convenience header that already includes most of the others provided in `roscpp`.

meyerj (commented):
Almost LGTM...

meyerj (approved):

PR #149, opened 2020-08-12
rtt_rosparam_tests: make Eigen optional according to rosparam.h

spd-intermodalics:
# Description

This PR watches changes the `rtt_rosparam_tests` test package for `rtt_rosparam` removing the assumption that the eigen support is always enabled, to make it depend on the definition:

The `rtt_rosparam` package offers a parameter `RTT_ROSPARAM_EIGEN_SUPPORT` to chooses the support with Eigen.
With this PR, the test also disables Eigen support if it is not supported in the original library.


meyerj (approved):
LGTM

PR #151, opened 2020-08-12
rtt_tf: add transform updates on Orocos ports

spd-intermodalics:
# Description

This PR provides support in `rtt_tf` to track TF transformations by publishing them periodically through an Orocos port using the `tf2_ros::TransformListener`.

## Contents

The next operations are added to the `rtt_tf::RTT_TF` Orocos component:
* `subscribeTransform(target, source)`: This operation generates a port of type `geometry_msgs::TransformStamped` dedicated to publish transformations between the two frames `target` and `source`.

## Getting updates

To get updates from the TF server, the `rtt_tf::RTT_TF` Orocos component needs to be running periodically.

## Test

The PR also updates the tests including the new feature.


rtt_tf/rtt_tf-component.cpp
```diff
@@ -206,10 +207,16 @@ namespace rtt_tf
     }
     // Publish the geometry messages obtained from the lookupTransform on
     // tracked transformations
-    for (const auto& tracker : ports_trackers) {
-      auto msg = lookupTransform(tracker.first.first, tracker.first.second);
-      tracker.second->write(msg);
-    }
+    // Range for not supported in ISO C++ 11
```
francisco-miguel-almeida: Range-based for loops were introduced in C++11.
Unless you actually meant C++98 in this comment?

rtt_tf/rtt_tf-component.cpp
```diff
@@ -80,7 +81,7 @@ namespace rtt_tf
   {
     tf2_msgs::TFMessage msg;
     // resolve names and copy transforms to message
-    msg.transforms.reserve(tforms.size());
+    msg.transforms.resize(tforms.size());
     std::transform(tforms.begin(), tforms.end(), msg.transforms.begin(), PrefixResolver(prefix));
```
francisco-miguel-almeida, line 85: You can do this in two ways here: either in this form, where the `std::transform` assumes the target container already has a matching size (requiring a call to `resize` above), or use `std::back_inserter(msg.transforms)` which results in transformed elements being `push_back`ed (which would not need `resize` above).

rtt_tf/rtt_tf-component.cpp
```diff
@@ -294,6 +332,48 @@ namespace rtt_tf
     port_tf_static_out.write(msg_out);
   }
 
+  bool RTT_TF::subscribeTransfrom(const std::string& target, const std::string& source)
+  {
+    Logger::In(this->getName());
+    const std::pair<std::string, std::string> transfrom_pair(target, source);
+    const std::string name_transform = source + std::string("_") + target;
```
francisco-miguel-almeida: `source + "_" + target` suffices here.

rtt_tf/rtt_tf-component.cpp
```diff
@@ -294,6 +332,48 @@ namespace rtt_tf
...
+    const std::string name_port = std::string("tracker_tf_") + name_transform;
+    if (ports_trackers.find(transfrom_pair) != ports_trackers.end()) {
+      RTT::log(RTT::Warning) << "The transfrom from " << source << " to " << target <<
+        " is already tracked." << RTT::endlog();
+      return false;
+    }
+    if (!canTransform(target, source)) {
```
francisco-miguel-almeida, line 346: Just to be clear: subscribing even though the transform is not yet available is the intentional behavior, correct?

francisco-miguel-almeida (approved):
Just a few small questions, but LGTM code wise.

rtt_tf/rtt_tf-component.cpp
```diff
@@ -196,6 +205,29 @@ namespace rtt_tf
...
+     }
+  }
+
+  bool RTT_TF::startHook()
+  {
+    if (0 == getPeriod()) {
+      RTT::log(RTT::Warning) << "The period of the component is 0 (zero), so no"
+        " updates form TF will be published automatically" << RTT::endlog();
+    }
+    buffer_core = BufferCorePtr(new tf2::BufferCore());
```
francisco-miguel-almeida: Initializing by calling `buffer_core.reset(new tf2::BufferCore())` instead is preferred.

rtt_tf/rtt_tf-component.cpp
spd-intermodalics: Yes, the transform could be broadcasted later because the producing component hasn't get operational yet, but trying to read is safe. If no transform exists, then this port will not output data.

rtt_tf/rtt_tf-component.cpp
spd-intermodalics: But if the elements are `push_back()`-ed into the `msg.transforms` vector, this would keep growing instead of substituting the elements in there, right? I don't clearly see the advantage.
This change was to solve the issue of `.reserve()` not actually initializing that memory and affecting `size()` but only the allocation and `capacity()` value. This caused a segfault before.

In other words, what happens if the new set of transforms is smaller in size than the previous one? Would it override only the first `tforms.size()` elements of `msg.transforms` or append to `msg.transforms` increasing its `size()` every time? What we want is that if fills an exact size of the vector.

francisco-miguel-almeida: I meant that one could use the `back_inserter` to `push_back()` elements after `reserve()`-ing instead of `resize()`-ing as you originally wrote. But indeed, all you would achieve is making sure that the heap allocation takes place at start, so there's no real advantage in this scenario.

francisco-miguel-almeida (approved):
LGTM.

PR #153, opened 2020-08-26
add acknowledgements to README.md

spd-intermodalics:
This PR adds a section to the `README.md` with the acknowledgements. The acknowledged list is taken from `rtt_ros2_integration`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,16 +329,14 @@ RTT components. See [rtt_actionlib](rtt_actionlib) for more information.
 
 ## Acknowledgement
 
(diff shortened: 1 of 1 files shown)


meyerj:
> The acknowledged list is taken from rtt_ros2_integration.

... but the text needs to be adapted then?

spd-intermodalics:
There was one paragraph indeed that wouldn't match well. I have fixed that.

dvanthienen (approved):

PR #156, opened 2020-09-11
rtt_roscomm: Shortcut ROS spinner callback queue for rtt subscribers

spd-intermodalics:
# Optimized ROS subscription pipeline using a custom `CallbackQueueInterface` implementation

Optimized ROS subscription pipeline, avoiding unnecessary copies by directly pushing the data to Orocos and bypassing unnecessary ROS threads.

## Goal

Do not depend on ROS spinner threads for subscriptions. Minimize latency of port reads and decouple multiple components with ROS subscriptions connected to ports. At the moment they all share the same pool of ROS spinner threads.

## Approach

The patch introduces a new "queue" without a queue to bypass the global `ros::CallbackQueue` of ROS for the subscripers of Orocos messages. Instead, the network thread will immediately execute the callback, and will pass it to the global `ros::CallbackQueue` only in case that it fails to execute with `ros::CallbackInterface::TryAgain` status.

## Implementation

* A new implementation of `ros::CallbackQueueInterface` is provided that doesn't hold a queue but directly calls the `call()` method of the `ros::CallbackInterface` received.
* Helper functions are provided to generate a subscriber with a custom implementation of `CallbackQueueInterface`.
* This method is chosen for the subscribers of the message transporter.

#### Callbacks and spinning
Explanation about Callbacks and Spinning:
http://wiki.ros.org/roscpp/Overview/Callbacks%20and%20Spinning

Check specially the section 4. Advanced: Using Different Callback Queues.
Our use-case is not in 4.1. Shortcutting the callback queue by not having a queue.

#### Function `Subscriber::subscribe()`

[Reference](https://github.com/ros/ros_comm/blob/ae2501fec33de9a19abe87199bbf60171c05dd34/clients/roscpp/include/ros/node_handle.h#L402)

All the overloaded `Subscriber::subscribe()` creates a `SubscribeOptions ops` and calls `subscribe(ops)`. The way to use a custom callback queue is to replace then entry in `SubscribeOptions` options as in `ops.callback_queue` different` from `0` (`0` is default `callback_queue`, the ROS global queue callback).

#### Function `CallbackQueue::callOneCB()`

[Reference](https://github.com/ros/ros_comm/blob/ae2501fec33de9a19abe87199bbf60171c05dd34/clients/roscpp/src/libros/callback_queue.cpp#L346)

The function `CallbackQueue::callOneCB()` executes a callback and then it removes it. The execution of the callback,
inherited from `CallbackInterface` is a simple `cb->call()`. In this case, the `CallbackInterface` is itself a `SubscriptionQueue` which can be executed through that `call()`.
The function also deals with `ros::CallbackInterface::TryAgain` result, which would be the case if in a multi-threaded situation the callback is being taken by another thread. We don't expect to get into this situation ever, but in case it happens, we can forward the callback to the global callback queue and let the ROS spinner deal with it, as in the original behavior.

#### Function `Subscription::handleMessage()`

[Reference](https://github.com/ros/ros_comm/blob/ae2501fec33de9a19abe87199bbf60171c05dd34/clients/roscpp/src/libros/subscription.cpp#L657)

The function `Subscription::handleMessage()` is the function that calls the `addCallback()` to push the callback to the global queue for a subscription.

#### Function `SubscriptionQueue::call()`

[Reference](https://github.com/ros/ros_comm/blob/121bac963e90b9a0bb33cc368ef4373a2f65ddff/clients/roscpp/src/libros/subscription_queue.cpp#L102)

The `SubscriptionQueue` is itself a `CallbackInterface` so it can be `call()`ed which executes normally all the needed delivery of the message.

### DIFF
--- a/rtt_roscomm/include/rtt_roscomm/passthrough_callback_queue.hpp
+++ b/rtt_roscomm/include/rtt_roscomm/passthrough_callback_queue.hpp
@@ -1,4 +1,3 @@
-
 #ifndef RTT_ROSCOMM__PASSTHROUGH_CALLBACK_QUEUE_HPP_
 #define RTT_ROSCOMM__PASSTHROUGH_CALLBACK_QUEUE_HPP_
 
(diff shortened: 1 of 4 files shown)


rtt_roscomm/include/rtt_roscomm/passthrough_callback_queue.hpp
```diff
@@ -0,0 +1,232 @@
+
```
meyerj, line 29: nit: remove empty line (and add a license header instead)

rtt_roscomm/include/rtt_roscomm/passthrough_callback_queue.hpp
```diff
@@ -0,0 +1,232 @@
...
+#ifndef RTT_ROSCOMM__PASSTHROUGH_CALLBACK_QUEUE_HPP_
+#define RTT_ROSCOMM__PASSTHROUGH_CALLBACK_QUEUE_HPP_
+
+#include <ros/ros.h>
+#include <ros/callback_queue_interface.h>
+#include <ros/subscription_queue.h>
+#include <ros/callback_queue.h>
+
+namespace rtt_roscomm {
+  class PassthroughCallbackQueue: public ros::CallbackQueueInterface
```
meyerj: style: no indentation inside `namespace {}` (but it is not consistent in existing files, too)

rtt_roscomm/include/rtt_roscomm/passthrough_callback_queue.hpp
```diff
@@ -0,0 +1,232 @@
...
+  {
+    public:
+      PassthroughCallbackQueue();
```
meyerj: It is trivial, so not needed to define a constructor explicitly at all.
If you do, you should probably also declare an explicit `virtual` destructor (the implicitly generated one is `virtual`, too).

rtt_roscomm/include/rtt_roscomm/passthrough_callback_queue.hpp
```diff
@@ -0,0 +1,232 @@
...
+
+      /** 
+       * Implementation of ros::CallbackQueueInterface::addCallback()
+       * 
+       * @param callback aaa
+       * @param owner_id aaa
+       * 
+       * @return void
+       */
```
meyerj: Either you should explain what this method is doing internally (i.e. invoke the callback directly), or omit the code documentation. Like this, it is pretty useless.

Same for `removeByID()`, which is a no-op. Could be implemented in the header to allow the compiler to inline it, and to document without the need of an additional comment that it is not doing anything here.

rtt_roscomm/include/rtt_roscomm/passthrough_callback_queue.hpp
```diff
@@ -0,0 +1,232 @@
...
+      virtual void removeByID(uint64_t owner_id);
+
+    private:
+
+      /**
+       * Port where to write the message
+       */
+      // base::PortInterface* output_port_ptr_;
+
```
meyerj:
```suggestion
```

remove unused code

rtt_roscomm/include/rtt_roscomm/rtt_rostopic_ros_msg_transporter.hpp
```diff
@@ -241,6 +242,7 @@ namespace rtt_roscomm {
     ros::NodeHandle ros_node;
     ros::NodeHandle ros_node_private;
     ros::Subscriber ros_sub;
+    PassthroughCallbackQueue prassthrough_callback_queue;
```
meyerj: typo:
```suggestion
    PassthroughCallbackQueue passthrough_callback_queue;
```

rtt_roscomm/include/rtt_roscomm/passthrough_callback_queue.hpp
```diff
@@ -0,0 +1,232 @@
...
+  }; // class PassthroughCallbackQueue
+
+  /**
+   * Helpers to generate subscribers with a custom ros::CallbackQueueInterface
+   */
```
meyerj: Not needed. This special implementation of `ros::CallbackQueueInterface` is used in exactly one place, in [`RosSubChannelElement::RosSubChannelElement()`](https://github.com/orocos/rtt_ros_integration/pull/156/files#diff-b7dd3168b5e27cadacbc7d8c54c72e8dR257) below. It should not be used anywhere else, because in most other contexts (with arbitrary user-defined callbacks) it won't do anything good.

The implementation of the constructor is already kind of unreadable, and using a `ros::SubscribeOptions ops` instance there would improve that anyway (see next comment). So all those helper functions should be removed. They are also not specific to `PassthroughCallbackQueue`, and last but not least would need to be maintained for future updates in the roscpp API that add or modify overloads.

Please see the suggested implementation of `RosSubChannelElement::RosSubChannelElement()` below.

rtt_roscomm/include/rtt_roscomm/rtt_rostopic_ros_msg_transporter.hpp
meyerj: Wrong order: The `PassthroughCallbackQueue` must exist before `ros::Subscriber`, such that it is destructed after the subscriber that uses it. Or the subscriber needs to be shutdown explicitly in the `~RosSubChannelElement()` destructor before the callback queue gets destroyed. But like this the subscriber refers to a dangling callback queue instance during destruction.

Probably calling `ros_sub.shutdown()` in the destructor would be safest.

rtt_roscomm/include/rtt_roscomm/rtt_rostopic_ros_msg_transporter.hpp
```diff
@@ -264,9 +266,9 @@ namespace rtt_roscomm {
         RTT::log(RTT::Debug)<<"Creating ROS subscriber for port "<<port->getName()<<" on topic "<<policy.name_id<<RTT::endlog();
       }
       if(topicname.length() > 1 && topicname.at(0) == '~') {
-        ros_sub = ros_node_private.subscribe(policy.name_id.substr(1), policy.size > 0 ? policy.size : 1, &RosSubChannelElement::newData, this); // minimum queue_size 1
+        ros_sub = subscribe(ros_node_private, &prassthrough_callback_queue, policy.name_id.substr(1), policy.size > 0 ? policy.size : 1, &RosSubChannelElement::newData, this); // minimum queue_size 1
       } else {
-        ros_sub = ros_node.subscribe(policy.name_id, policy.size > 0 ? policy.size : 1, &RosSubChannelElement::newData, this); // minimum queue_size 1
+        ros_sub = subscribe(ros_node, &prassthrough_callback_queue, policy.name_id, policy.size > 0 ? policy.size : 1, &RosSubChannelElement::newData, this); // minimum queue_size 1
       }
     }
```
meyerj, line 279: Suggestion without the need for any of the `rtt_roscomm::subscribe()` template methods defined in `passthrough_callback_queue.hpp`:

```cpp
    RosSubChannelElement(RTT::base::PortInterface* port, const RTT::ConnPolicy& policy) :
      ros_node(),
      ros_node_private("~")
    {
      ros::SubscribeOptions ops;
      ops.initByFullCallbackType<T>(
          policy.name_id,
          policy.size > 0 ? policy.size : 1,  // minimum queue_size 1
          boost::bind(&RosSubChannelElement::newData, this, _1));
      ops.callback_queue = &passthrogh_callback_queue;

      RTT::Logger::In in(policy.name_id);
      if (port->getInterface() && port->getInterface()->getOwner()) {
        RTT::log(RTT::Debug) << "Creating ROS subscriber for port " << port->getInterface()->getOwner()->getName() << "." << port->getName() << " on topic " << ops.topic << RTT::endlog();
      } else {
        RTT::log(RTT::Debug) << "Creating ROS subscriber for port " << port->getName() << " on topic " << ops.topic << RTT::endlog();
      }

      if (ops.topic.length() > 1 && ops.topic[0] == '~') {
        ops.topic = ops.topic.substr(1);
        ros_sub = ros_node_private.subscribe(ops);
      } else {
        ros_sub = ros_node.subscribe(ops);
      }
    }
```

rtt_roscomm/src/passthrough_callback_queue.cpp
```diff
@@ -0,0 +1,24 @@
+
+#include "rtt_roscomm/passthrough_callback_queue.hpp"
+
+namespace rtt_roscomm {
+
+PassthroughCallbackQueue::PassthroughCallbackQueue()
+        {}
+
```
meyerj:
```suggestion
```

rtt_roscomm/src/passthrough_callback_queue.cpp
```diff
@@ -0,0 +1,24 @@
...
+void PassthroughCallbackQueue::addCallback(
+    const ros::CallbackInterfacePtr &callback,
+    uint64_t owner_id)
+{
+  // Call CallbackInterface::CallResult SubscriptionQueue::call()
+  if (ros::CallbackInterface::TryAgain == callback->call()) {
```
meyerj: nit: Make the call more explicit here? For example
```suggestion
  const ros::CallbackInterface::CallResult result = callback->call();
  if (ros::CallbackInterface::TryAgain == result) {
```

That would be obvious without the need for the additional comment.

rtt_roscomm/src/passthrough_callback_queue.cpp
```diff
@@ -0,0 +1,24 @@
+
```
meyerj, line 29: nit: remove empty line and add a license header instead

meyerj (changes requested):

rtt_roscomm/include/rtt_roscomm/passthrough_callback_queue.hpp
spd-intermodalics: Not adding a license header, otherwise is more problematic to maintain. Instead, the file inherits the `LICENSE` in the root of the package (or project).

rtt_roscomm/src/passthrough_callback_queue.cpp
spd-intermodalics: Same as before, I prefer not to add the license header.

rtt_roscomm/include/rtt_roscomm/passthrough_callback_queue.hpp
```diff
@@ -8,224 +7,39 @@
...
-  }
+class PassthroughCallbackQueue: public ros::CallbackQueueInterface
+{
+  public:
+    /** 
+     * Implementation of ros::CallbackQueueInterface::addCallback()
+     * 
+     * This method is executing the callback received instead of adding
+     * it to a queue. In this way, the queue is bypassed and immediately
+     * executed.
```
meyerj:
```suggestion
     * it to a queue. In this way, the queue is bypassed and the callback is immediately
     * executed.
```

But actually that sentence is just repeating the first.

rtt_roscomm/include/rtt_roscomm/passthrough_callback_queue.hpp
```diff
@@ -8,224 +7,39 @@
...
+     * 
+     * @param callback callback to execute, instead of queueing
+     * @param owner_id Owner of the callback, in other implementations it might
+     *                 be used to remove the callback. In this implementation
+     *                 it has no effect.
```
meyerj:
```suggestion
     * @param owner_id unused
```

nit: You do not need to explain what other implementations of `ros::CallbackQueueInterface` would probably do with that parameter, or how it is intended to be used by the base class. That is already [documented by roscpp](http://docs.ros.org/api/roscpp/html/classros_1_1CallbackQueueInterface.html) itself.

rtt_roscomm/include/rtt_roscomm/passthrough_callback_queue.hpp
```diff
@@ -8,224 +7,39 @@
...
+     * 
+     * @return void
```
meyerj:
```suggestion
```

rtt_roscomm/include/rtt_roscomm/passthrough_callback_queue.hpp
```diff
@@ -8,224 +7,39 @@
...
+        uint64_t owner_id=0);
+
+    /** 
+     * Implementation of ros::CallbackQueueInterface::removeByID()
+     * 
+     * No-op
+     * 
+     * @param owner_id Not used.
+     * 
+     * @return void
```
meyerj:
```suggestion
```

PR #162, opened 2020-12-09
rtt_rosparam: add general support for sequence property serialization and deserialization

### LINKED ISSUE #155: Add the possibility to unmarshall rosparam array to sequence property
If I define a sequence property type that is not hard-coded into rtt_rosparam, as, for instance, `std::vector<std::pair<double, double>>`, it is not possible to unmmarshall it from rosparam, even if I declare the type in a typekit.

If I have a "array" XmlRpc item in the rosparam, the property is not unmarshalled (pseudo-code).
```
# Before unmarshall
my_property = { {0.0, 0.0} }
/my/property: [{first: 0.1, second: 0.2}, {first: 0.3, second: 0.4}]
# After unmarshall
my_property == { {0.0, 0.0} }
```

If I use a struct XmlRpc item like:
```
# Before unmarshall
my_property = { {0.0, 0.0} }
/my/property/Element0/first: 0.1
/my/property/Element0/second: 0.2
/my/property/Element1/first: 0.3
/my/property/Element1/second: 0.3
# After unmarshall
my_property == { {0.1, 0.2} }
```

only the pairs that are already allocated in my property's vector are updated, but the vector is neither extended nor shrunk.

Hugal31:
For get() (ROS -> Orocos), add a special case to deconstruct array XmlRpcValue into property
  bags, fill it with "ElementX", as Orocos does itself.
For set() (Orocos -> ROS), detect if the property is a sequence by looking at a write-only "size" int property, deconstruct it in a PropertyBag, and read its "ElementX" member properties, as [RTT::types::propertyDecomposition](https://github.com/orocos-toolchain/rtt/blob/600102e8be9c81905b20930e32d43b28244ab173/rtt/types/PropertyDecomposition.cpp#L158) does.

Add the corresponding unit tests.

Closes #155.

### DIFF
--- a/rtt_rosparam/src/rtt_rosparam_service.cpp
+++ b/rtt_rosparam/src/rtt_rosparam_service.cpp
@@ -1,8 +1,10 @@
 #include <rtt/RTT.hpp>
 #include <rtt/Property.hpp>
+#include <rtt/internal/DataSource.hpp>
 #include <rtt/plugin/ServicePlugin.hpp>
 #include <rtt/types/PropertyDecomposition.hpp>
 
+#include <boost/lexical_cast.hpp>
 #include <boost/utility/enable_if.hpp>
 #include <boost/type_traits/is_convertible.hpp>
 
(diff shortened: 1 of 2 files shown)


rtt_rosparam/src/rtt_rosparam_service.cpp
```diff
@@ -432,6 +434,28 @@ XmlRpc::XmlRpcValue rttPropertyToXmlParam(const Eigen::VectorXf &vec)
   return xml_array;
 }
 
+// We assume that a property with an unsassignable "size" int member is a sequence.
```
meyerj: nit: typo
```suggestion
// We assume that a property with an unassignable "size" int member is a sequence.
```

meyerj (approved):
One typo, but apart of that LGTM! Thanks for your contribution!

Hugal31:
I have some troubles making this work on the pipeline. This is probably because I am using ROS noetic, and not ROS kinetic as the pipeline. I'll try if I can reproduce a kinetic environment.

Hugal31:
Is it possible to merge this? It is ready now.

meyerj:
> Is it possible to merge this? It is ready now.

Yes, sorry for the delay!

PR #168, opened 2021-11-02
rtt_rosparam: add rosparam.getParams(ns) function

### LINKED ISSUE #167: Add a way to get the rosparams to populate the root service
For now, there is no way to populate the "root service" of a TaskContext with a custom ros param name. But it is possible with a sub service:

```
rosparam set /bar/int 42
rosparam set /bar/string "forty-two"
```

```
loadComponent("foo", "Components::Foo")
loadService("foo", "rosparam")
// Will populate foo.bar properties
foo.rosparam.getParam("/bar", "bar")
// Will fail with [ Debug  ][ROSParamService::getParam] RTT component does not have a property or service named ""
foo.rosparam.getParam("/bar", "")
```

There is no way to populate "foo" with properties from "/bar". This could be easily achieved by adding for example a check for `""` [here](https://github.com/orocos/rtt_ros_integration/blob/df79afb7a7b6e90ea7b807f0e4322662caee8c07/rtt_rosparam/src/rtt_rosparam_service.cpp#L924).

Hugal31:
Allows to populate the root service with a specific non-leaf ROS parameter.

I am not sure if I should do this PR on toolchain-2.9 or 2.10.

Closes #167.

### DIFF
--- a/rtt_rosparam/src/rtt_rosparam_service.cpp
+++ b/rtt_rosparam/src/rtt_rosparam_service.cpp
@@ -108,7 +108,7 @@ public:
       .arg("param_name", "Name of the ROS parameter. Use '~' and '/' leaders for private or absolute resolution.")
       .arg("name", "Name of the RTT property or service.");
     this->addOperation("getParams", static_cast<bool(ROSParamService::*)(const std::string&)>(&ROSParamService::getParams), this)
-      .doc("Populates the properties of this component from the ROS param server based on the given ROS parameter name.")
+      .doc("Populates the properties of this component from the ROS param server based on the given ROS parameter namespace.")
       .arg("param_name", "Name of the ROS parameter. Use '~' and '/' leaders for private or absolute resolution.");
 
     this->addOperation("getRelative", &ROSParamService::getParamRelative, this)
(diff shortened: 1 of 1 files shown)


rtt_rosparam/src/rtt_rosparam_service.cpp
```diff
@@ -107,6 +107,9 @@ class ROSParamService: public RTT::Service
       .doc("Gets one property of this component (or populates the properties of a named RTT sub-service) from the ROS param server based on the given ROS parameter name.")
       .arg("param_name", "Name of the ROS parameter. Use '~' and '/' leaders for private or absolute resolution.")
       .arg("name", "Name of the RTT property or service.");
+    this->addOperation("getParams", static_cast<bool(ROSParamService::*)(const std::string&)>(&ROSParamService::getParams), this)
+      .doc("Populates the properties of this component from the ROS param server based on the given ROS parameter name.")
```
meyerj: nit:
```suggestion
      .doc("Populates the properties of this component from the ROS param server based on the given ROS parameter namespace.")
```

meyerj (approved):
LGTM, thanks! Sorry for the late review and reaction.

PR #174, opened 2024-06-12
rtt_roscomm: generate dependency typekits

### LINKED ISSUE #173: rtt_roscomm: There is no dependency handling when generating typekits
If a ROS message `my_pkg_1/MyMsg1` depends on a message `my_pkg_2/MyMsg2` from another package and the typekit for the first package is generated using `ros_generate_rtt_typekit(my_pkg_1)`, then, when attempting to use use this type in Orocos, this results in a compilation error due to the `boost::serialization::serialize()` function for `my_pkg_2/MyMsg2` not being defined.

It would be nice if typekits could also be generated automatically for these types of dependencies, to avoid having to hunt down and list every single dependency recursively in order to have complex ROS messages work.

I found a workaround by generating typekits for all dependencies listed in `${${my_pkg_1}_MSG_DEPENDENCIES}`, which is set by calling `include(${${my_pkg_1}_PREFIX}/${CATKIN_GLOBAL_SHARE_DESTINATION}/${my_pkg_1}/cmake/${my_pkg_1}-msg-paths.cmake)`. I believe something similar could be done.

danieto98:
* When generating typekits, also generate them for all packages that the current package depends on for msg generation.
* Pass dependencies onto targets correctly so they are generated in the right order.
* Don't generate typekits or service proxies if they have already been generated (if there's a corresponding target or if the corresponding rtt_pkg already exists).

Closes #173

### DIFF
--- a/rtt_roscomm/cmake/GenerateRTTROSCommPackage.cmake.em
+++ b/rtt_roscomm/cmake/GenerateRTTROSCommPackage.cmake.em
@@ -43,12 +43,68 @@ endmacro()
 
 macro(ros_generate_rtt_typekit package)
   cmake_parse_arguments(ros_generate_rtt_typekit "" "" "EXTRA_INCLUDES" ${ARGN})
-  set(_package ${package})
-  add_subdirectory(${rtt_roscomm_TEMPLATES_DIR}/typekit ${package}_typekit)
+
+  find_package(rtt_${package} QUIET)
+
+  if(NOT rtt_${package}_FOUND AND NOT TARGET rtt-${package}-typekit)
+    set(_include_dirs)
+    set(_pkg_libs)
+    set(_msg_exported_targets)
+    set(_typekit_targets)
+    set(_transport_targets)
+    set(_mqueue_targets)
+    set(_corba_targets)
+
+    include(${${package}_PREFIX}/${CATKIN_GLOBAL_SHARE_DESTINATION}/${package}/cmake/${package}-msg-paths.cmake)
+
+    foreach(dep ${${package}_MSG_DEPENDENCIES})
+      find_package(${dep} QUIET)
+
+      if(${dep}_FOUND)
+        list(APPEND _include_dirs ${${dep}_INCLUDE_DIRS})
+        list(APPEND _pkg_libs ${${dep}_LIBRARIES})
+        list(APPEND _msg_exported_targets ${${dep}_EXPORTED_TARGETS})
+
+        find_package(rtt_${dep} QUIET)
+
+        if(NOT rtt_${dep}_FOUND AND NOT TARGET rtt-${dep}-typekit)
+          ros_generate_rtt_typekit(${dep})
+        endif()
+        
+        if(rtt_${dep}_FOUND)
+          list(APPEND _pkg_libs ${rtt_${dep}_LIBRARIES})
+        endif()
+
+        if(TARGET rtt-${dep}-typekit)
+          list(APPEND _typekit_targets rtt-${dep}-typekit)
+        endif()
+
+        if(TARGET rtt-${dep}-ros-transport)
+          list(APPEND _transport_targets rtt-${dep}-ros-transport)
+        endif()
+
+        if(TARGET rtt-${dep}-ros-transport-mqueue)
+          list(APPEND _mqueue_targets rtt-${dep}-ros-transport-mqueue)
+        endif()
+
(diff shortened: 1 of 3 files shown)


rtt_roscomm/cmake/GenerateRTTROSCommPackage.cmake.em
```diff
@@ -43,12 +43,68 @@ endmacro()
 
 macro(ros_generate_rtt_typekit package)
   cmake_parse_arguments(ros_generate_rtt_typekit "" "" "EXTRA_INCLUDES" ${ARGN})
-  set(_package ${package})
-  add_subdirectory(${rtt_roscomm_TEMPLATES_DIR}/typekit ${package}_typekit)
+
+  find_package(rtt_${package} QUIET)
```
meyerj, line 47: This approach would prevent that the typekit is rebuilt in an existing workspace after it has been built once, for example after one message definition was updated?

meyerj (changes requested):
Thanks for your contribution and for suggesting a patch already to resolve #173!

The way `rtt_roscomm` was intended to be used by the original authors was with the [`create_rtt_msgs`](https://github.com/orocos/rtt_ros_integration/blob/d58546d81152898e21efe79f400c164e5d944c90/rtt_roscomm/scripts/create_rtt_msgs) script as documented [here](https://github.com/orocos/rtt_ros_integration/tree/d58546d81152898e21efe79f400c164e5d944c90/rtt_roscomm#making-a-packages-ros-msg-and-srv-types-available), and one `rtt_foo_msgs` package per `foo_msgs` package, where the `rtt_bar_msgs` created by that script also declares a dependency on `rtt_foo_msgs` if `bar_msgs` depends on `foo_msgs`. So the dependency handling is in fact available, but not in CMake code. The typekit packages must be created manually once, one per ROS package with message and/or service definitions. The build tool (e.g. catkin) is then able to build the typekit packages in the correct order, following their build dependencies topology. Updated dependencies of the underlying message packages would then require a recreation of the respective typekit package, too, or manual updates.

I am not sure anymore, but I don't think any build time dependencies are required at all, also not for generated headers? So the generated Boost header for `bar_msgs` does not include the generated header for `foo_msgs`, nor need the resulting libraries to be linked to each other. As a consequence, the `rtt_foo_msgs` and `rtt_bag_msgs` packages can be built in any order and do not depend on the artifacts of the other. Or not at all. Only at run-time, when loaded into an Orocos process, certain features may not function properly if no typekit is available for nested types.

In my opinion we should not allow a call to macro `ros_generate_rtt_typekit(<package>)` that may implicitly includes typekit or service proxy plugins for other packages, only because they cannot be found at the time CMake is invoked. A catkin workspace can be built multiple times and should support selective rebuilding if only one file, for example a message definition, changes. With your suggested change, the outcome of a successful build would be that afterwards all `rtt_${package}` exist, and hence after another CMake invocation without a full clean the targets will effectively be disabled and never updated anymore?

As an alternative to the "one package - one typekit package" paradigm implied by [`create_rtt_msgs`](https://github.com/orocos/rtt_ros_integration/blob/d58546d81152898e21efe79f400c164e5d944c90/rtt_roscomm/scripts/create_rtt_msgs), what *should* work is to manually create a typekit package for a collection of maybe related ROS message packages, by calling the `ros_generate_rtt_typekit(<package>)` macro multiple times in the same package-level `CMakeLists.txt` file. At least that did work once with `catkin_make` and a single build directory for all packages in a workspace, which effectively results in a single top-level `CMakeLists.txt` at workspace level that descends into packages like subdirectories in their dependency order. Maybe some `add_dependencies()` calls are missing at the moment indeed to support those macro calls in any order within a single package, such that each typekit library targets has a dependency on the `*-generate_boost_headers` targets for each of its dependencies, even if empty in this package.

What could be done and accepted for being merged is an additional macro with the behavior intended by this pull request, like `ros_generate_rtt_typekits_recursively(<package>)`, but without changing `ros_generate_rtt_typekit(<package>)`. Still, the criteria when to abort the recursion should be more explicit in my opinion, and cannot depend on whether the workspace was already built once before or not.

Another flaw with this approach is that if a typekit library for `foo_msgs` is provided by a package `rtt_bar_msgs`, because `bar_msgs` dependes on `foo_msgs` and package `rtt_foo_msgs` is not found, then another package's typekit library that also depends on `foo_msgs` would again include the `rtt_foo_msgs` library, because `find_package(rtt_foo_msgs)` would still fail and there is no such package.

